PR #210, opened 2015-01-08
Support for multiple fonts

markusscherer:
This implements the possibility of using more than one font for nice typographical effects. Use as described in `readme.md`.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -198,7 +198,7 @@ For the output template:
 - `<fc=#FF0000>string</fc>` will print `string` with `#FF0000` color
   (red).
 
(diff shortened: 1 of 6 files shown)


jaor:
This looks pretty good! Just one more nit: i know it was me who proposed
the name `fontList`, but looking at how this list is used in conjunction
with `font` i was thinking that perhaps a better name for the config
file field would be `additionalFonts`.  What do you think?

markusscherer:
I agree, `additionalFonts` captures the semantics better.

jaor:
excellent: merged! i've also added you to the credits.

PR #212, opened 2015-01-31
Added additional wind parser case for "Variable at"

dino-:
Wind data sometimes looks like this:

  "Wind: Variable at 3 MPH (3 KT):0"

Note the missing direction info. This fix adds an additional
parser for this case to pWind.


jaor:
looks good: merged (after a rebase). thanks!

PR #220, opened 2015-03-16
New Monitor plugin: Showing UV rate

romanofski:
Hi,

I live in Australia. One thing we're wary off is the sun and the bodies exposure to it. That's why I wrote this plugin to show the UV data from the Australian Radiation Protection agency. The plugin grabs the exported XML from the agency and updates it's value in an interval.

I'm not sure about any coding conventions and mind you, I've a beginner to Haskell. Please feel free to let me know if this Plugin makes sense to be included or not.
Due to the limited audience this plugin might serve, I've made it optional. I've also updated the readme in order to run this plugin.

If there are any code changes violating current style or more changes need to be made to include this, let me please know.

Thanks for the great software! Kind Regards.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -157,6 +157,9 @@ Otherwise, you'll need to install them yourself.
 :    Support for xpm image file format. This will allow loading .xpm files in `<icon>`.
      Requires the [libXpm] C library.
 
+`with_uvmeter`
+:    Enables UVMeter plugin. The plugin shows UV data for Australia.
+
 `all_extensions`
 :    Enables all the extensions above.
 
@@ -1242,6 +1245,20 @@ more than one battery.
 - Example:
 
     Run CatInt 0 "/sys/devices/platform/thinkpad_hwmon/fan1_input" [] 50
+
+### `UVMeter`
+
+- Aliases to "uv " + station id. For example: `%uv brisbane%` or `%uv
+  alice springs%`
+- Args: default monitor arguments.
+
+- *Reminder:* Keep the refresh rate high, to avoid making unnecessary
+  requests every time the plug-in is run.
+- Station IDs can be found here:
+  http://www.arpansa.gov.au/uvindex/realtime/xml/uvvalues.xml
+- Example:
+
+        Run UVMeter "brisbane" ["-H", "3", "-L", "3", "--low", "green", "--high", "red"] 900
 
 ## Executing External Commands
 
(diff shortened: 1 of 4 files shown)


jaor:
@romanofski my apologies for the long time to get back to you.  the code looks ready to merge, thanks a lot.  would you mind squash all commits to one and rebase on top of current master? if you prefer to keep the history of your commits for some reason let me know, and we can also merge this way.

romanofski:
Updated the push request and squashed everything in one commit. Should I update also auxiliary files such as readme.md and news.md ?

One other thing: I think it would be a good idea to add extra caution to the patch. My language skills are at beginner level and I hope I'm not slipping in foolish bugs with this patch.

Cheers!

jaor:
@romanofski the code looks fine and we can merge this.  but you might want to consider re-implementing the plugin using Parsec, like the Weather plugin, to avoid introducing the rather heavy-weight dependency on conduit, which has lots of dependencies.  if you prefer not to do that, i can live with that since it's an optional feature... what do you think?

romanofski:
I can imagine that it is a pretty big dependency for an optional plugin which will only be useful for a small amount of people. Btw. It'll be cool if I could find more feeds from other stations to make it more generic.

I'd like to give Parsec a try. If I fail miserable we can still keep this patch :)

Cheers!

jaor:
excellent, thanks roman! and yeah, in the end we can always keep this
implementation since it's optional (perhaps not including it under the
all_extensions catch-all, so that only requesting it explicitly works).
agreed also on making it more generic.

romanofski:
In a14ebbc02589f04751d1e3c0393434d14ef13932 I have an implementation using Parsec. This basically adds an additional XML parser to the plug in. I think it makes sense for one plug in, if there are more plug ins who need an XML parser that would obviously not suffice. The code is now very specific to the format the Australian agency exports, however I'm not aware of any other agency which publishes it's data.

Let me know if that code goes in the right direction. Haven't tested it yet, that's something for the weekend :)

jaor:
yes, this is indeed what i had in mind when i suggested using parsec.
and since the plugin doesn't introduce any new dependencies, we could
even have it compiled by default (perhaps we could set the flag to true
by default and keep it for hard-core minimalists), or, at any rate,
activate it whe all_extensions is set.

i say, go for it! :)

romanofski:
Ok. After testing I've ironed out one annoying thing in 31944a2c5c38214d94c8acc802b1d5a5bc4b57d9 to avoid that an HTTP error clutters the whole bar.  Other than that I'm happy with the change. If you are as well, squash and update I assume?

romanofski:
I took the liberty of updating and squashing the pull request. If there are any changes to be made I think I can do it on the updated/squashed change as equally as to two separate patches :)

jaor:
merged! thanks a lot

romanofski:
W00h00! Thanks again for the great program.

PR #230, opened 2015-05-03
Explanatory warning when xft font is requsted but not supported

ahgilbert:
Hi, I spent longer than I should have fiddling with font settings. It turned out that I had failed to build xmobar with the with_xft flag set. I figured I'd save the next person who doesn't read the docs closely enough from the same trouble.

### DIFF
--- a/src/XUtil.hsc
+++ b/src/XUtil.hsc
@@ -33,6 +33,7 @@ import Control.Concurrent
 import Control.Monad (when)
 import Control.Monad.Trans
 import Control.Exception (SomeException, handle)
+import Data.List
 import Foreign
 import Graphics.X11.Xlib hiding (textExtents, textWidth)
 import qualified Graphics.X11.Xlib as Xlib (textExtents, textWidth)
@@ -49,7 +50,6 @@ import qualified System.IO as UTF8 (readFile,hGetLine)
 # endif
 #endif
 #if defined XFT
-import Data.List
 import MinXft
 import Graphics.X11.Xrender
 #endif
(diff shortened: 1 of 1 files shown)


src/XUtil.hsc
```diff
@@ -81,12 +81,18 @@ data XFont = Core FontStruct
 -- to the Xft backend Example: 'xft:Sans-10'
 initFont :: Display ->String -> IO XFont
 initFont d s =
-#ifdef XFT
        let xftPrefix = "xft:" in
        if  xftPrefix `isPrefixOf` s then
+#ifdef XFT
```
jaor: the two let's above are supposed to be only seen whe XFT is defined...

jaor:
this is a very good idea, but you left out a couple of lines in your implementation and compiling with xft gives me the warning too...  thanks a lot!

ahgilbert:
On my system (32 bit debian, ghc 7.10, cabal 1.22), I need to do a 'cabal clean' between building without the flag, and building with it. Otherwise, I see the warning even after building with the flag set. Could you give that a try? I believe that the logic is correct.

Sorry if you receive this message twice--I replied by email, but do not see my comment on the pull request.

src/XUtil.hsc
ahgilbert: The 'let' and 'if' are necessary to determine if the user is trying to use an xft font. The preprocessor flag now only controls what xmobar's reaction is--using the xft font, or emitting a warning.

jaor:
you're right (also regarding those two lines): pushed!

PR #240, opened 2015-08-18
Switch Weather Plugint to Network.HTTP.Conduit

dragosboca:
Network.HTTP does not honor the proxy settings while Network.HTTP.Conduit does.

This will help users running behind a proxy.


jaor:
this looks useful, but using conduits almost doubles the number of dependencies of xmobar... i guess there's no leaner alternative? or we could perhaps put the code under a conditional flag...

dragosboca:
Ok. 
I have defined a new flag (with_conduit)

jaor:
Excellent.  Merged, thanks!

PR #244, opened 2015-12-09
Add template variables for wind speed in m/s and km/h to weather plugin

berndbrot:
I'm quite new to Haskell, so a thorough review is recommended.

### DIFF
--- a/src/Plugins/Monitors/Weather.hs
+++ b/src/Plugins/Monitors/Weather.hs
@@ -100,6 +100,10 @@ noWind = WindInfo "μ" "μ" "0" "0" "0" "0"
 pWind :: Parser WindInfo
 pWind =
   let tospace = manyTill anyChar (char ' ')
+      toKmh knots = knots $* 1.852
+      toMs knots  = knots $* 0.514
+      ($*) :: String -> Double -> String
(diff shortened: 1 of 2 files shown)


src/Plugins/Monitors/Weather.hs
```diff
@@ -115,8 +119,13 @@ pWind =
...
                 knot <- tospace
                 manyTill anyChar newline
-                return $ WindInfo cardinal azimuth mph knot
+                return $ WindInfo cardinal azimuth mph knot (toKmh knot) (toMs knot)
   in try wind0 <|> try windVar <|> try wind <|> return noWind
+  where
+    toKmh knots = knots $* 1.852
+    toMs knots  = knots $* 0.514
+    ($*) :: String -> Double -> String
+    op1 $* op2 = show (round ((read op1::Double) * op2)::Integer)
```
jaor: a nit: you could put the definitions in this `where` clause in the `let` (for instance, just below `toSpace`), and then one only has to look for defs in one place...

berndbrot:
Thanks for the hint, I moved the functions accordingly.

jaor:
merged, thanks a lot!

PR #245, opened 2015-12-15
New Message Printing Plugin

andrew-m-h:
A plugin designed to read an inspirational message from a file (new line seperated) and print it out at regular intervals. I personally use it with a file of 3000 "fun facts" that print out every 10 seconds because I use 2 xmobar's (top & bottom) and Have just enough free space on the bottom for a couple of 100 character's of fun fact.

### DIFF
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -43,6 +43,8 @@ import Plugins.MBox
 
 import Plugins.DateZone
 
+import Plugins.InspMsg
+
 -- $config
 -- Configuration data type and default configuration
 
@@ -150,5 +152,6 @@ infixr :*:
 runnableTypes :: Command :*: Monitors :*: Date :*: PipeReader :*: BufferedPipeReader :*: CommandReader :*: StdinReader :*: XMonadLog :*: EWMH :*: Kbd :*: Locks :*:
                  Mail :*: MBox :*:
                  DateZone :*: MarqueePipeReader :*:
+                 InspMsg :*: 
                  ()
 runnableTypes = undefined
--- a/src/Plugins/InspMsg.hs
+++ b/src/Plugins/InspMsg.hs
@@ -9,10 +9,13 @@ fun facts, historical quotes...
 
 module Plugins.InspMsg (InspMsg(..)) where
 
-import Plugins (Exec(..), readFileSafe, tenthSeconds)
+import Plugins (Exec(..), tenthSeconds)
 
-import Prelude (String, Int, IO, Show(..), Read(..),
-               lines, length, ($))
+import Prelude (Int, IO, String, Show(..), Read(..),
+               length, ($))
+--Using Text and Text.IO reduces overhead significantly for large files
+import Data.Text (Text, unpack, lines) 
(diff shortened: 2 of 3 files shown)


edvorg:
Why not use Com plugin?

andrew-m-h:
Well, my first solution was to re-purpose https://github.com/jgoerzen/ledmon to read  and use the Com plugin. But as it stands, the Com plugin doesn't seem to be appropriate to read and randomize the data. I also like to have a native plugin where possible.

edvorg:
But it's quite easy to do it with Com like `cat file | sort -R | head -1`

andrew-m-h:
I must admit, it never occurred to me. However, one reason may be efficiency. sort isn't lazy, neither is cat and so for a very large file. it could be an unreasonably high overhead. This is fixed with this plugin since the file is read in once into an array. I'm not entirely sure what the overhead of cat file | sort -R | head -1 is (on very large files) but it can't be better than the plugin.

edvorg:
I understand your point, however it's never occurred to me also :)

andrew-m-h:
Interestingly enough, I've done some testing on a rather large file (200520 lines) and the overhead for cat file | sort -R | head -1 was 1.8 seconds (on my computer) which is significant. A similar result was obtained for my plugin. So I've changed the plugin to use Data.Text instead of Strings to hold the data. This improved the plugin about 50-100 times.

edvorg:
What do you think if make it more flexible and provide two configurable options:
1) the source we get the line from (file, web request, remote file)
2) provide the rule - how we select an item from source (pick random item, pick latest item, etc.)
So it will be useful in many interesting ways and not only in reading random things from local dictionary.

edvorg:
E.g. I can monitor some stuff from remote machine just buy getting file over ssh and showing some content from it using this plugin.

andrew-m-h:
You interest me. I'll do some research and see what I can do. Though I agree that the idea(s) are very good. Feel free to contribute if you've any knowledge on the Network.HTTP library and such like.

jaor:
i like this idea

rethab:
I am at least a little concerned about the feasibility of such a generic plugin. 

Displaying a random line from a file is entirely different from pulling stuff from the web and then displaying something meaningful (hardly any serious webservice will give you data where you can say I care about line x or so. What you usually get is JSON or some format, which would require more involved parsing (good luck with integrating that as well).

Then there are a ton of strategies of how you may want to pull the data. While you can read a file once and then keep it in memory and display random lines once in a while, in the case of a webservice you'd probably want to call the service again, etc.

Don't get me wrong, I have thought about somehow integrating stuff from the web into my status bar as well, however, that generic idea diverges quite a bit from the original ideas. Both of which - quite frankly - could be achieved using the Com approach. In particular the idea with the file over SSH will mostly be shell commands anyway unless you want to integrate file-over-ssh in the plugin as well (in which case my two points from above are even stronger ;-) )

I am curious to see with what you come up.

\- Reto

jaor:
moved to https://codeberg.org/jao/xmobar/pulls/245

PR #250, opened 2015-12-30
Added "-p" launch option

regellosigkeitsaxiom:
`-p "string"` will override position settings set in configuration file.
Useful for launching multiple xmobars from single configuration file,
which is almost necessary when using non-monospace fonts.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -394,6 +394,7 @@ xmobar --help):
       -A alpha      --alpha=alpha          The transparency: 0 is transparent, 255 (the default) is opaque
       -o            --top                  Place xmobar at the top of the screen
       -b            --bottom               Place xmobar at the bottom of the screen
+      -p            --position=position    Specificty position, same as in config file
       -d            --dock                 Try to start xmobar as a dock
       -a alignsep   --alignsep=alignsep    Separators for left, center and right text
                                            alignment. Default: '}{'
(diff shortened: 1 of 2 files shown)


jaor:
this looks good, but could you please also document the new option in readme.md? thanks!

regellosigkeitsaxiom:
Done.

readme.md
```diff
@@ -394,6 +394,7 @@ xmobar --help):
       -A alpha      --alpha=alpha          The transparency: 0 is transparent, 255 (the default) is opaque
       -o            --top                  Place xmobar at the top of the screen
       -b            --bottom               Place xmobar at the bottom of the screen
+      -p            --position=position    Specificty position, same as in config file
```
jaor: typo: Specify

src/Main.hs
```diff
@@ -38,6 +38,8 @@ import System.Environment
 import System.FilePath ((</>))
 import System.Posix.Files
 import Control.Monad (unless, liftM)
+import Text.Read (readMaybe)
+import Data.Maybe (fromMaybe)
```
jaor: this import seems redundant? i don't see any use of `fromMaybe`...

regellosigkeitsaxiom:
Yes, you are right.

jaor:
excellent: i've squashed all your commits to a single one and pushed it. thanks a lot!

PR #255, opened 2016-01-13
change the interval time for check pipe

zlbruce:
`threadDelay` is in microseconds, It is unnecessary to check pipe every millisecond


jaor:
thanks!

PR #260, opened 2016-03-01
feat(MPD): add <date> support in template

Delapouite:
Hi.

With this patch, users can now display the release date of their favorite songs playing on MPD.


jaor:
thanks!

PR #268, opened 2016-07-16
Add expandEnv function and use it in PipeReader

incertia:
expandEnv takes a string and expands the environment variables it can
find. variable substringing (e.g. ${VAR:1} to lop off the first
character) is not supported, but $VAR and ${VAR} formats are, with the
former being delimited by punctuation, but not underscores.

ideally, we should be able to do a one time expansion in all strings, but I do not know where to stick the function call.

this is useful, especially for pipereader, if you want to have a similar xmobar configuration across multiple x displays (say you are running a vnc server), and want to avoid race conditions by specifying a different pipe (say xmobar.pipe.${DISPLAY:1}) for each display.

### DIFF
--- a/src/Environment.hs
+++ b/src/Environment.hs
@@ -1,0 +1,48 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMobar.Environment
+-- Copyright   :  (c) William Song
+-- License     :  BSD-style (see LICENSE)
+--
+-- Maintainer  :  Will Song <incertia@incertia.net>
+-- Stability   :  stable
+-- Portability :  portable
+--
+-- A function to expand environment variables in strings
+--
+-----------------------------------------------------------------------------
+module Environment where
+
+import Data.Maybe           (fromMaybe)
+import System.Environment   (lookupEnv)
+
+expandEnv :: String -> IO String
+expandEnv "" = return ""
+expandEnv (c:s) = case c of
+  '$'       -> do
+    envVar <- fromMaybe "" <$> lookupEnv e
+    remainder <- expandEnv s'
+    return $ envVar ++ remainder
+    where (e, s') = getVar s
+          getVar "" = ("", "")
+          getVar ('{':s) = (takeUntil "}" s, drop 1 . dropUntil "}" $ s)
+          getVar s = (takeUntil filterstr s, dropUntil filterstr s)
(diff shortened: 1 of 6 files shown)


incertia:
I took a look at Parsers.hs. It appears as if the most logical place to stick expandEnv would be [L80](https://github.com/jaor/xmobar/blob/master/src/Parsers.hs#L80), because it is not very clear to me how config strings get converted into instances of Runnable and so on.

src/Environment.hs
```diff
@@ -0,0 +1,48 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMobar.Environment
+-- Copyright   :  ???
```
jaor: (c) <your name here>

src/Environment.hs
```diff
@@ -0,0 +1,48 @@
...
+-- License     :  BSD-style (see LICENSE)
+--
+-- Maintainer  :  ???
```
jaor: if you don't feel like putting here your name, mine will do (`jao <jaog@gnu.org>`)

src/Environment.hs
```diff
@@ -0,0 +1,48 @@
+-----------------------------------------------------------------------------
```
jaor: this module needs to be added to the list in `xmobar.cabal`... i think you'll find the spot easily enough.

jaor:
@incertia this looks mostly good... yes it'd be nice to have automatic expansion everywhere, but we can start with PipeReader and monitor-specific.  I've made a couple of comments to complete the patch.  Also: would you mind adding a documentation bit to readme.md (in PioeReader's section) for the new capability?  Thanks a lot!

incertia:
made the changes you requested, as well as sticking the module in the two other pipe readers for the sake of consistency

jaor:
excellent! would you mind squashing all the commits to one on top of current master? or, if you prefer to keep your history of changes, at least rebase them? thanks a lot!

incertia:
is something like the above good (first time squashing for a PR)?

jaor:
yeah, the squashing went fine! i've rebased the resulting commit and merged it to master, followed by a couple of silly fixes for compilation warnings and an update or the readme to include your name as a contributor. thanks! (closing manually, just discard your branch and resync your master)

PR #281, opened 2016-09-14
tell the user what went wrong in the event of a failure

cleverca22:
with this patch, it will now give more info on what is failing when things go wrong, for example:

``Thread checker failed: thread blocked indefinitely in an STM transaction``


jaor:
merged and pushed. thanks!

PR #284, opened 2016-11-04
Fix MultiCpu monitor for large uptimes

zevweiss:
This is essentially commit a0397ad2 applied to MultiCpu.


jaor:
merged, thanks!

PR #289, opened 2017-01-23
Fix Net monitor for large uptimes/bytecounts

liskin:
My laptop currently has rx/tx bytes in 10s of gigabytes and it's only been up
for 20 days. Normally it's several times more. At this point, Float can only
tell the difference of 4KB and up:

Prelude> let x = (50 * 2^30 :: Float) in (x + 2000) - x 0.0 Prelude> let x =
(50 * 2^30 :: Float) in (x + 3000) - x 4096.0

This commit makes the Net monitor read Word64 which is exactly what the kernel
prints into /proc/net/dev [1] and converts to Float only after subtracting the
two numbers.

[1]
https://github.com/torvalds/linux/blob/7a308bb3016f57e5be11a677d15b821536419d36/net/core/net-procfs.c#L82

Still, I think it's time to switch from Float to Double. At half-gigabit 
speeds (easily attainable at home while rsyncing over a direct UTP cable 
between two post-2010 laptops), Float can only tell the difference of 8 bytes
and up (and I'm not even considering takeDigits!). That's probably okay for a
Net monitor in xmobar, but we're so close to the limit it makes sense to move
to Double just in case.

### DIFF
--- a/src/Plugins/Monitors/Net.hs
+++ b/src/Plugins/Monitors/Net.hs
@@ -20,6 +20,7 @@ module Plugins.Monitors.Net (
 
 import Plugins.Monitors.Common
 
+import Data.Word (Word64)
 import Data.IORef (IORef, newIORef, readIORef, writeIORef)
 import Data.Time.Clock (UTCTime, getCurrentTime, diffUTCTime)
 import Control.Monad (forM, filterM, liftM)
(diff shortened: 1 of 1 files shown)


liskin:
Related: https://github.com/jaor/xmobar/commit/35ca0c36f251003a2123433035b6f3a838cff4a4#commitcomment-20574078
(mentioning it here since github comments often get lost)

jaor:
looks good: merged.  and yes, i think your reasoning is sound and moving to Double make sense.

liskin:
Ok, I'll try to prepare another with Double and also fix Cpu/MultiCpu to Word64.

jaor:
excellent, very much appreciated Tomáš!

PR #304, opened 2017-05-29
readme: specify package name for iwlib on Arch

b0o:
On Arch Linux, the package providing `iwlib` is a bit hard to find - it
should come from the official `wireless_tools` package [1]. I've updated
the readme to make this more clear.

[1] https://www.archlinux.org/packages/?q=wireless_tools


jaor:
thanks, @b0o. i've also upated the website over at http://xmobar.org

PR #305, opened 2017-06-30
fixed a typo

parvmor:


jaor:
@parvmor hm, the option is actually `m`, or so the code says: what makes you think it's an `n`?

parvmor:
The [code](https://github.com/jaor/xmobar/blob/master/src/Plugins/Monitors/Common.hs#L177) says that normal color's short flag is `n`. `m` is for minWidth.

jaor:
@parvmor that's not the same flag.  the flag you mention is documented [here](https://github.com/jaor/xmobar/blob/master/readme.md#default-monitor-arguments).

For the Battery monitor there are *additional*  flags that come after the special "--" marker. `-m` is one of them, and it's implemented [here](https://github.com/jaor/xmobar/blob/master/src/Plugins/Monitors/Batt.hs#L71).  There's an example of its usage just a bit below the lines your PR touches.

parvmor:
Ahh okay. I might have misplaced `-m` while configuring the rc.

PR #307, opened 2017-07-30
Remove UVMeter from global 'other-modules' list

michaellilanushoober:
List the module UVMeter within the conditional branch of the flag 'with_uvmeter' instead of the global 'other-modules'.


jaor:
thanks!

PR #309, opened 2017-07-30
Guard monitor 'Weather' behind flag 'with_weather'

michaellilanushoober:
Hide the monitor 'Weather' behind a flag named 'with_weather'.
This implies that the dependency on the module 'HTTP' is now optional.

Further moved the effect of the flag with_conduit within the branch of
'with_weather', since that flag only has some effect if 'with_weather'
is true.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -160,9 +160,13 @@ Otherwise, you'll need to install them yourself.
 `with_uvmeter`
 :    Enables UVMeter plugin. The plugin shows UV data for Australia.
 
+`with_weather`
+:    Support to display weather information. Enables Weather plugin.
+
 `with_conduit`
 :    Use http-conduit for getting weather data enabling support for http proxies.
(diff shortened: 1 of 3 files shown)


jaor:
hmmm, i'm a bit hesitant here.  the weather plugin has been always built in by default and not considered an extension.  so users that simply build xmobar without specifying any flag expect the weather plugin to be there: perhaps we could set the default of with_weather to True?

michaellilanushoober:
Thanks for the feedback. Setting with_weather to True is fine with me. I've updated the branch accordingly.

jaor:
Thanks Michael.  After merging your previous PR this branch does not merge cleanly anymore with master.  Would you mind rebasing it, please? Thanks!

michaellilanushoober:
Done. While I was at I, I also took the liberty to squash the two commits into one.

jaor:
excellent. thanks!

PR #312, opened 2017-08-24
Fix UVMeter due to changed URL and BOM

romanofski:
The URL which exports the real time UV data has changed to an HTTPS
address. Since the HTTP package does not support HTTPS URLs, use
http-conduit to retrieve the XML document.

Unfortunately, the XML documents XML declaration precedes a byte order
mark which the previous XML parser was unable to handle. We're simply
ignoring the BOM in order to get to the UV values.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -158,7 +158,8 @@ Otherwise, you'll need to install them yourself.
      Requires the [libXpm] C library.
 
 `with_uvmeter`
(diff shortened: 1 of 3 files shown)


jaor:
looks good and travis is happy (even though github doesn't seem to know, i wonder why): merged. thanks!

PR #331, opened 2018-01-30
Fixing a dependency error with dbus

sahabi:
Versions of dbus > 0.13 are problematic when using cabal to build the project.

PR #332, opened 2018-01-30
Fix travis/CI

sahabi:
The Travis CI fails due to the command sh ./travis.sh src returns an
error due to the fact that travis.sh does not exist.

The fix is removing the command from travis.yml.


sahabi:
Please note this PR fails the CI checks due the issue of dbus dependency conflict. The fix for that is in this [PR](https://github.com/jaor/xmobar/pull/331).

PR #335, opened 2018-02-03
Change behaviour of CpuFreq plugin

Fulmene:
1. Make GHz value honour -d (decimal digits) switch since it is a decimal point value.
2. If -S (suffix) if false, the number should be of the same unit, so I think it should display MHz value.

### DIFF
--- a/src/Plugins/Monitors/CpuFreq.hs
+++ b/src/Plugins/Monitors/CpuFreq.hs
@@ -35,8 +35,9 @@ runCpuFreq _ = do
   ddigits <- getConfigValue decDigits
   let path = ["/sys/devices/system/cpu/cpu", "/cpufreq/scaling_cur_freq"]
       divisor = 1e6 :: Double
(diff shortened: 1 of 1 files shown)


jaor:
thanks for your patch. i am not sure it's a good idea to let the use of a
suffix decide the units we use: what if someone (e. g. me :)) wants to use
Gz without any suffix? perhaps what we need is a different flag dictating
the unit to use?

Fulmene:
Well, my main gripe is just that the number isn't always of the same unit, so it suddenly goes from 1.xx - 3.xx to hundreds. Personally I'm fine with defaulting to GHz too. :)

jaor:
oh, then let's default to Gz and we don't have to introduce any new flags.
would you like to adapt this pr to make it so?

jaor:
thank you, @Fulmene .  merged, and mentioned in readme.md.

jaor:
(btw @Fulmene, i added your name to the readme.md file, hope that's okay)

Fulmene:
It's OK :)

PR #339, opened 2018-03-03
Add ellipsis option for Monitor's fields and total text

o-schneider:
1. Add global ellipsis option for both Monitor's fields and total texts 
2. Add tests infrastructure based on hspec and tests on the new option
3. Add IntelliJ projects files to the gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,7 @@
 cabal.sandbox.config
 .stack-work
 /web/readme.md
+/.idea/
+/xmobar.iml
+/out/
+/cabal.config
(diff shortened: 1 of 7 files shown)


jaor:
thanks olivier, this looks good (and i really appreciate your
introducing unit tests, something i should have done long, long ago).
would you mind to squash the commits to one, so that we avoid the
biggish cabal.config also in git's history. thanks again!

o-schneider:
Glad you appreciate the PR ! There will certainly be some more coming, that's why I integrated unit tests. ;) . The squash is done and pushed.

PR #340, opened 2018-03-17
Add fallback icon support to MultiCpu

bobtwinkles:
To make the `<autoipat>` pattern truly automatic, there needs to be a fallback
icon set to use when the user hasn't specified enough `--load-icon-patterns`.
This adds the fallback under the name `--fallback-icon-pattern`


jaor:
merged, thanks!

PR #341, opened 2018-03-20
Split Package Into A Library & Executable

prikhi:
Adding a library component lets other Haskell programs generate Config values & run xmobar.

I've structured it so most of the code is in the library, with the executable as a small wrapper that parses the config & cli arguments, does some initialization, & runs the loop.

I moved the Main module into a separate `app/` folder, and nested every other module under the `Xmobar` module. Which meant I had to prepend  `Xmobar.` to every module header & local import. It makes the PR diff huge, I suggest viewing the diff of each commit instead of the whole PR.

I then did the most controversial bit of this PR, I replaced the cabal file with an hpack `package.yaml` file. Stack uses this to generate the cabal file. It adds a `library` section which exposes everything, some global dependencies, ghc options, & conditionals. 

I understand that it's a big, possibly unwanted change & I can revert it to use a cabal file instead, it just made it a little easier for me.

The last 2 commits fix some bugs the package.yaml exposed. I only tested building with the default flags, with the all_extensions flag, &  with all the default flags disabled.

I use this to define xmobar configs & launch xmobar in xmonad. E.g., I can use colors from my `Theme` module in my Config, and compose configurations:

https://github.com/prikhi/xmonad-config/blob/dc8ddf7c67d050c058297c7086da39fcaab7dae8/src/StatusBar.hs#L142-L197 

I do still require some additional code to get xmobar running, but I wasn't sure if that code should live in this repo, or xmonad-contrib, or just my personal config.

I lifted the initialization code from the Main module(minus arg/config parsing) & ran it with forkProcess:

https://github.com/prikhi/xmonad-config/blob/dc8ddf7c67d050c058297c7086da39fcaab7dae8/src/XmobarStub.hs#L48-L74

And probably less related for this repo, I created a named pipe & added a pipereader to the config commands so I could send the workspace status to the forked process:

https://github.com/prikhi/xmonad-config/blob/dc8ddf7c67d050c058297c7086da39fcaab7dae8/src/XmobarStub.hs#L23-L45

Lemme know what you think!


prikhi:
Ah, the travis script directly uses cabal instead of stack... I can update it if we'll keep the package.yaml file, otherwise it should just keep using cabal.

package.yaml
```diff
@@ -0,0 +1,240 @@
...
+    - changelog.md
+    - samples/xmobar.config
+    - samples/xmonadpropwrite.hs
+    - samples/Plugins/helloworld.config
+    - samples/Plugins/HelloWorld.hs
+
+
+dependencies:
+    - base >= 4.0 && < 5
+    - X11 >= 1.6.1
```
prikhi: I kept all the version bounds from the cabal file, but they're probably unnecessary w/ stack?

package.yaml
```diff
@@ -0,0 +1,240 @@
...
+library:
+    source-dirs: src
+    extra-libraries:
+        - Xrandr
+        - Xrender
+    dependencies:
+        - base >= 4.0 && < 5
+        - mtl >= 2.1 && < 2.3
+        - parsec == 3.1.*
+        - stm >= 2.3 && < 2.5
```
prikhi: Same here,  does this override the stack LTS version?

src/Xmobar/Config.hs
```diff
@@ -152,7 +157,12 @@ infixr :*:
...
 -- this function's type signature.
 runnableTypes :: Command :*: Monitors :*: Date :*: PipeReader :*: BufferedPipeReader :*: CommandReader :*: StdinReader :*: XMonadLog :*: EWMH :*: Kbd :*: Locks :*:
+#ifdef INOTIFY
                  Mail :*: MBox :*:
-                 DateZone :*: MarqueePipeReader :*:
+#endif
+#ifdef DATEZONE
+                 DateZone :*:
+#endif
+                 MarqueePipeReader :*:
```
prikhi: I wanted to put MarqueePipeReader after `Locks :*:` but wasn't sure if ordering mattered here.

jaor:
@prikhi sorry for the huge delay... i still plan to look at this! thanks for your patience

prikhi:
@jaor All good, take your time. I'll fix the merge conflicts once you get a chance to review it.

PR #347, opened 2018-04-05
Net: fix race condition in netParser/readNetDev/isUp

liskin:
It's possible (and happens) that netParser sees a device that is no
longer there when we get to isUp. This happens almost every time I
resume from suspend and the WWAN card (USB device) reappears, and
results in xmobar showing the exception until I restart it.

Originally I tried

    tryJust (guard . isDoesNotExistError) (B.readFile (operstateDir d))

and it worked for a while but in recent kernels it's possible to open
the file and have the device disappear before we get to reading the
contents of it, so we need to surround the whole open/read block in
catchIOError and make sure it's evaluated.


jaor:
merged, thanks a lot!

Issue #346, opened 2018-04-05
Doesn't build with dbus 1.0.1


felixonmars:
```
[ 9 of 57] Compiling IPC.DBus         ( src/IPC/DBus.hs, dist/build/xmobar/xmobar-tmp/IPC/DBus.dyn_o )

src/IPC/DBus.hs:46:32: error:
    Ambiguous occurrence ‘interfaceName’
    It could refer to either ‘DBus.Client.interfaceName’,
                             imported from ‘DBus.Client’ at src/IPC/DBus.hs:18:1-18
                          or ‘IPC.DBus.interfaceName’, defined at src/IPC/DBus.hs:33:1
   |
46 | sendSignalMethod mvst = method interfaceName sendSignalName
   |                                ^^^^^^^^^^^^^
```

Hiding the name brings more errors:

```
[ 9 of 57] Compiling IPC.DBus         ( src/IPC/DBus.hs, dist/build/xmobar/xmobar-tmp/IPC/DBus.dyn_o )

src/IPC/DBus.hs:43:34: error:
    • Couldn't match expected type ‘Interface’
                  with actual type ‘[Method]’
    • In the third argument of ‘export’, namely
        ‘[sendSignalMethod mvst]’
      In a stmt of a 'do' block:
        export client objectPath [sendSignalMethod mvst]
      In the expression:
        do client <- connectSession
           requestName client busName [nameDoNotQueue]
           export client objectPath [sendSignalMethod mvst]
   |
43 |         export client objectPath [ sendSignalMethod mvst ]
   |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^

src/IPC/DBus.hs:46:25: error:
    • Variable not in scope:
        method
          :: InterfaceName
             -> MemberName
             -> Signature
             -> Signature
             -> (MethodCall -> IO Reply)
             -> Method
    • Perhaps you meant data constructor ‘Method’ (imported from DBus.Client)
   |
46 | sendSignalMethod mvst = method interfaceName sendSignalName
   |                         ^^^^^^

src/IPC/DBus.hs:58:18: error:
    • Variable not in scope: replyReturn :: [a0] -> Reply
    • Perhaps you meant data constructor ‘ReplyReturn’ (imported from DBus.Client)
   |
58 |         return ( replyReturn [] )
   |                  ^^^^^^^^^^^
```

jaor:
what version of xmobar are you trying? in the current xmobar.cabal we
[...]

PR #352, opened 2018-05-24
Fixes for dbus 1.0

ijaketak:
for #346

### DIFF
--- a/src/IPC/DBus.hs
+++ b/src/IPC/DBus.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  DBus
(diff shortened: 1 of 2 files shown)


jaor:
@ijaketak thanks a lot, this looks mostly good.  the only glitch is that the build fails for ghc 7.8: seems that dbus 1.0 doesn't even compile for that ghc 7.8.  if so, perhaps we could use preprocessor conditionals, althought it sounds kind of messy... what do you think?

peti:
Ping? Please release this fix soon'ish. Support for ghc 7.8.x really sounds like it's not *that* important, to be honest.

jaor:
On Tue, Jul 10 2018, Peter Simons wrote:

> Ping? Please release this fix soon'ish. Support for ghc 7.8.x really
> sounds like it's not that important, to be honest.

Well, that depends on how the maintainers in several linux distributions feel
about it.  An option here is do the work to maintain both versions, if
this is urgent.

jaor:
Sorry, sent too quickly... Personally, i have no real interest in ghc 7.8, but i wouldn't like to break other people's builds... also, i am not really sure what dbus-1.0 brings to the table for us over previous versions: is it fixing something important?

peti:
dbus-1.x is part of both Stackage LTS 12.x and Nightly, which is the standard many distributions such as NixOS and openSUSE follow. If xmobar does not support it, then it's impossible to build the package in those environments.

jaor:
@peti Many thanks for the clarification.  I've pushed a PR similar to this one that makes xmobar use dbus 1.x and drops compatibility with GHC 7.x.  @ijaketak thanks a lot for the PR! Closing manually.

peti:
Thank you very much for the new release!

PR #353, opened 2018-06-04
port to dbus-1

trofi:
Signed-off-by: Sergei Trofimovich <slyfox@gentoo.org>


jaor:
We just dropped support for GHC 7.x, and i cherry-picked your commit above (so closing the PR manually).  We now support dbus > 1.0.  Thanks!

PR #358, opened 2018-07-23
Fix cabal and ghc warnings

osa1:


jaor:
hi. those code changes break compilation, and the list of modules you're
adding is already there... perhaps you were using an outdated source?

osa1:
It turns out the variable is conditionally used, but the modules are used even when the flags are not set (or at least that's what cabal thinks). Here's the warnings I'm getting:

```
src/Xmobar.hs:286:11: warning: [-Wunused-matches]
    Defined but not used: ‘wr’
    |
286 | drawInWin wr@(Rectangle _ _ wid ht) ~[left,center,right] = do
    |           ^^
[51 of 51] Compiling Main             ( src/Main.hs, /home/omer/haskell/xmobar_2/dist-newstyle/build/x86_64-linux/ghc-8.4.2/xmobar-0.27/x/xmobar/build/xmobar/xmobar-tmp/Main.o )

<no location info>: warning: [-Wmissing-home-modules]
    These modules are needed for compilation but not listed in your .cabal file's other-modules: Plugins.DateZone
                                                                                                 Plugins.MBox
                                                                                                 Plugins.Mail
```

osa1:
Closing this as properly fixing these will take more time.

Issue #290, opened 2017-01-27
Force redraw


d125q:
Is there a way to force xmobar to redraw whenever, e.g., the volume is changed?  I am thinking of a signal or something.

jaor:
> Is there a way to force xmobar to redraw whenever, e.g., the volume is
> changed? I am thinking of a signal or something.

xmobar reacts to SIGUSR signals:
http://projects.haskell.org/xmobar/#signal-handling

so one way woud be to have a script that changes volume and then sends
SIGUSR2 to xmobar.  that should work without modifying xmobar at all.

is that what you need?

d125q:
Yes, that is what I want, but sending `SIGUSR2` to xmobar does not cause it to redraw.

jaor:
then we'll have to mark this as a feature request :) although i'm not
sure if the fact that it doesn't redraw was intended or is just an
oversight.  i'll take a look as soon as i can (unless someone else beats
me, of course!)

jaor:
another possibility: do you use dbus? if so, sending the Wakeup command
might work (using "string:Wakeup" in
http://projects.haskell.org/xmobar/#the-dbus-interface)

skinner33:
I think the issue is not with the redraw, but with the volume plugin's inner workings:
The plugins polls regularly the current volume. Thus it does not update instantly on a volume change. Forcing a redraw will not change that.

I tried, i think about a year ago, to change that, but i did not found any callback/event-handler/... interface for pulseaudio to make the volume plugin event-based.

The workaround @d125q could use is to increase the refreshrate of the volume plugin.
Off the the top of my head one could also (ab)use stdinreader for displaying the volume, although some hacking would be required. Or try the message plugin for which a pull-request is open.

A more permanent fix within xmobar could be to change the way plugins are handled: Instead of using sleeps in each plugin thread, just use one main thread to schedule the different refresh rates and then unblock the relevant plugins (via a STM). I had this idea floating around for some time, but never came around to make a proof of concept. Initially i want to try this to reduce the number of wake ups.

mitchellwrosen:
The `"string:Wakeup"` command does not cause a re-draw. Is it supposed to?

Issue #328, opened 2018-01-12
CommandReader's program isn't cleaned up on exit 


mkoskar:
Using `CommandReader` plugin it seems that when `xmobar` exits the command keeps running re-parented under pid1. I think these background commands should be cleaned up / terminated on exit. As a workaround I wrap xmobar in a simple script:

```
#!/bin/bash

cleanup() { trap : TERM; kill 0; }
trap cleanup EXIT

xmobar
```

jaor:
moved to https://codeberg.org/jao/xmobar/issues/328

PR #360, opened 2018-08-25
Add event-based refreshing for Volume plugin (--monitor option)

DanielSchuessler:
This uses `alsactl monitor` for push/event-based updating instead of polling. Unless users encounter problems with this, I'd suggest making it the default in the future, since it reacts quicker and more smoothly, and avoids wasting cycles when nothing has changed :)

Side note: Duplicating the `if flag(...) ...` stuff in the test-suite stanza of the .cabal file is of course ugly. Maybe there should be an internal library, with the exe (stub) and the test-suite depending on it?

Fixes (kind-of) #290.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1063,6 +1063,14 @@ more than one battery.
     - `--highd` _number_ High threshold for dB. Defaults to -5.0.
     - `--lowd` _number_ Low threshold for dB. Defaults to -30.0.
     - `--volume-icon-pattern` _string_ dynamic string for current volume in `volumeipat`.
+    - `--monitor[=/path/to/alsactl]`
+        - Use event-based refreshing via `alsactl monitor` instead of polling
+          (`RefreshRate` will be ignored).
+        - If no `/path/to/alsactl` is given, `alsactl` will be sought in your `PATH`
+          first, and failing that, at `/usr/sbin/alsactl` (this is its location on
+          Debian systems. `alsactl monitor` works as a non-root user despite living
+          in `/usr/sbin`.).
+        - `stdbuf` (from coreutils) must be (and most probably already is) in your `PATH`.
 - Variables that can be used with the `-t`/`--template` argument:
             `volume`, `volumebar`, `volumevbar`, `volumeipat`, `dB`, `status`
 - Note that `dB` might only return 0 on your system. This is known
--- a/readme.md
+++ b/readme.md
@@ -1063,14 +1063,6 @@ more than one battery.
     - `--highd` _number_ High threshold for dB. Defaults to -5.0.
     - `--lowd` _number_ Low threshold for dB. Defaults to -30.0.
     - `--volume-icon-pattern` _string_ dynamic string for current volume in `volumeipat`.
-    - `--monitor[=/path/to/alsactl]`
-        - Use event-based refreshing via `alsactl monitor` instead of polling
-          (`RefreshRate` will be ignored).
-        - If no `/path/to/alsactl` is given, `alsactl` will be sought in your `PATH`
-          first, and failing that, at `/usr/sbin/alsactl` (this is its location on
-          Debian systems. `alsactl monitor` works as a non-root user despite living
-          in `/usr/sbin`.).
-        - `stdbuf` (from coreutils) must be (and most probably already is) in your `PATH`.
 - Variables that can be used with the `-t`/`--template` argument:
             `volume`, `volumebar`, `volumevbar`, `volumeipat`, `dB`, `status`
 - Note that `dB` might only return 0 on your system. This is known
@@ -1079,6 +1071,21 @@ more than one battery.
 - Requires the package [alsa-core] and [alsa-mixer] installed in your
   system. In addition, to activate this plugin you must pass
   `--flags="with_alsa"` during compilation.
+
+### `Alsa Mixer Element Args`
+
+Like [Volume](#volume-mixer-element-args-refreshrate), but with the following differences:
+- Uses event-based refreshing via `alsactl monitor` instead of polling, so it will refresh
+  instantly when there's a volume change, and won't use CPU until a change happens.
+- Aliases to `alsa:` followed by the mixer name and element name separated by a colon. Thus,
+  `Alsa "default" "Master" []` can be used as `%alsa:default:Master%`.
+- Additional options (after the `--`):
+    - `--alsactl=/path/to/alsactl`
+        - If this option is not specified, `alsactl` will be sought in your `PATH`
+          first, and failing that, at `/usr/sbin/alsactl` (this is its location on
+          Debian systems. `alsactl monitor` works as a non-root user despite living
+          in `/usr/sbin`.).
+- `stdbuf` (from coreutils) must be (and most probably already is) in your `PATH`.
 
 ### `MPD Args RefreshRate`
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /dist/
+/dist-newstyle/
 /TAGS
 /IWlib.hs
 /StatFS.hs
@@ -17,3 +18,5 @@ cabal.sandbox.config
 /xmobar.iml
 /out/
 /cabal.config
+codex.tags
+cabal.project.local
--- a/src/Main.hs
+++ b/src/Main.hs
@@ -38,18 +38,18 @@ import System.Exit
 import System.Environment
 import System.FilePath ((</>))
 import System.Posix.Files
-import Control.Exception
 import Control.Concurrent.Async (Async, cancel)
+import Control.Exception (bracket)
 import Control.Monad (unless)
 import Text.Read (readMaybe)
 
-import Signal (setupSignalHandler)
+import Signal (setupSignalHandler, withDeferSignals)
 
 -- $main
 
 -- | The main entry point
 main :: IO ()
-main = do
+main = withDeferSignals $ do
   initThreads
   d <- openDisplay ""
   args <- getArgs
@@ -77,6 +77,7 @@ main = do
 
 cleanupThreads :: [[([Async ()], a)]] -> IO ()
 cleanupThreads vars =
+  -- putStrLn "In cleanupThreads"
   for_ (concat vars) $ \(asyncs, _) ->
     for_ asyncs cancel
 
--- a/src/Main.hs
+++ b/src/Main.hs
@@ -25,6 +25,7 @@ import Parsers
 import Config
 import XUtil
 
+import Data.Foldable (for_)
 import Data.List (intercalate)
 import qualified Data.Map as Map
 
@@ -37,6 +38,8 @@ import System.Exit
 import System.Environment
 import System.FilePath ((</>))
 import System.Posix.Files
+import Control.Exception
+import Control.Concurrent.Async (Async, cancel)
 import Control.Monad (unless)
 import Text.Read (readMaybe)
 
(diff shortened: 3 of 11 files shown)


jaor:
@DanielSchuessler this looks nice.  I was thinking that instead of having to pass a command line flag to the existing `Volume`, we could perhaps define a new monitor, say `VolumeMonitor` that takes, instead o a polling time,  the path to alsactl as argument in the config file.  That way, it's always clear when one is using one or the other, and one keeps all the params on the config file.

Totally agree on your side note.  We have this nifty PR #341 open that implements xmobar as a library, but unfortunately I took so long to review it that its author seems (very understandably!) to have lost interest (now it needs to be rebased, etc.): recovering that PR should also address the code repetition issue you mention.

DanielSchuessler:
@jaor: It does work from the config file, e.g. in my xmobarrc:

```hs
Run Volume "default" "Master"
            [ "-t", "<dB> dB<status>"
            , "--"
            , "--on", ""
            , "--off", " [off]"
            , "--monitor"
            ]
            10
```

Regarding making a new `VolumeMonitor` plugin: Wouldn't that make it impossible to make the `--monitor` mode the default in the future (I don't really see a reason to use polling mode unless bugs are found with the `--monitor` mode)? (Well, not impossible, but removing `VolumeMonitor` again would be a breaking change).

About #341: It looks like you are only unsure about the part that replaces the `.cabal` with a `package.yaml`. I think the other changes could be applied without the `package.yaml` part?

jaor:
On Sun, Aug 26 2018, DanielSchuessler wrote:

> @jaor: It does work from the config file, e.g. in my xmobarrc:
>
> Run Volume "default" "Master"
>             [ "-t", "<dB> dB<status>"
>             , "--"
>             , "--on", ""
>             , "--off", " [off]"
>             , "--monitor"
>             ]
>             10

Ah! Sorry, I totally misread the code.

> Regarding making a new VolumeMonitor plugin: Wouldn't that make it
> impossible to make the --monitor mode the default in the future (I
> don't really see a reason to use polling mode unless bugs are found
> with the --monitor mode)? (Well, not impossible, but removing
> VolumeMonitor again would be a breaking change).

Well, we could always keep Volume simply as an alias of VolumeMonitor,
but I see what you mean.  What bothers me a bit is keeping the poll
period parameter as mandatory when the default is monitoring: it's ugly,
no?  In an ideal world, we would make the parameter optional, but I
don't think we can.

If we kept VolumeMonitor (or some better name) along with Volume, we
would at least both of them use only require parameters, and we could
advertise the new one as the recommended one, with a simpler
syntax... maybe we could name it `Alsa`?

     Run Alsa "default" "Master"
              [ "-t", "<dB> dB<status>"
              , "--"
              , "--on", ""
              , "--off", " [off]"
              --- this is optional: , "--monitor", "some non-std path"
             ] -- no polling frequency here

Just brainstorming here... let me know your thoughts...


> About #341: It looks like you are only unsure about the part that
> replaces the .cabal with a package.yaml. I think the other changes
> could be applied without the package.yaml part?

Probably, but I haven't looked closey at the merge conflicts.  And yes,
I am still not totally convinced to switch to stack after using it for a
little while, so ideally we'd do it in two phases.

DanielSchuessler:
Amended as suggested :)

DanielSchuessler:
Yes, I think it's a questionable hlint suggestion because the list will have to be re-added as soon as there's more than one option (same with "redundant do"s in the hspec tree that I had already removed), but *shrug*.

jaor:
Hmm, we have another problem: when xmobar exists, the alsactl process is not killed, so restarting xmobar causes alsactl zombies to accumulate...  I don't know off-hand how to accomplish that with Haskell's system interface...

readme.md
```diff
@@ -1068,6 +1068,21 @@ more than one battery.
...
 
+### `Alsa Mixer Element Args`
+
+Like [Volume](#volume-mixer-element-args-refreshrate), but with the following differences:
+- Uses event-based refreshing via `alsactl monitor` instead of polling, so it will refresh
+  instantly when there's a volume change, and won't use CPU until a change happens.
+- Aliases to `alsa:` followed by the mixer name and element name separated by a colon. Thus,
+  `Alsa "default" "Master" []` can be used as `%alsa:default:Master%`.
+- Additional options (after the `--`):
+    - `--alsactl=/path/to/alsactl`
```
jaor: hmm, i cannot make this option work... the usual way would be

             ["--", "-a", "path/to"]

or

             ["--", "--alsactl", "/path/to"]

DanielSchuessler: Oh right, `startAlsaPlugin` delegates to `Volume.runVolume`, which uses only its own `parseOpts`. Looking at `runVolume`, is there a reason why it reparses its options every time it's called, rather than parsing them just once? (The `args` list doesn't seem to ever change, looking at `runMBD` and `doArgs`. This might actually be relevant to all Monitor plugins.)

DanielSchuessler:
Sorry about the zombies. The usage of [withCreateProcess](http://hackage.haskell.org/package/process-1.6.4.0/docs/System-Process.html#v:withCreateProcess) was supposed to handle this, but apparently it never reaches its cleanup function.

I just tried moving the `withCreateProcess` scope from the reader thread to the main thread of the plugin, to no avail. Do the plugin threads ever get terminated by raising an exception on them, so they can run their `finally`/`bracket` blocks? If not, I think this would be necessary, but it would be a serious change... (The `async` package would be handy).

Alternatively, is there some way to ensure GHC cleanly terminates all background threads when the program exits?

DanielSchuessler:
Update: This is harder than expected. After I got the main thread to signal termination to, and wait for, all the plugin threads when the main thread exits, it cleaned up correctly if `xmobar` is SIGINTed. But not with SIGTERM, SIGQUIT etc., which kill the process immediately by default (apparently GHC installs no handlers for these). So we'd have to catch all these signals and make them `throwTo` the main thread.

On the positive side, the lingering `alsactls` aren't actual zombies; they are adopted by `init` and they seem to terminate upon the next volume change due to the broken pipe.

jaor:
Yes, I faced the same problems in a first attempt at fixing #328 that
didn't pan out.  I solution suggested somewhere (I've forgotten where)
was to make sure that the child processes belong to the same group as
xmobar's, and there was some way in System of doing that, albeit in a
kind of convoluted way.

DanielSchuessler:
So it finally seems to work with all the signal handlers, a bit drastic though :) I went through `man 7 signal` and took all the `Term` and `Core` signals that aren't already handled (there are also the Real-Time signals, but I think they aren't supported by `System.Posix.Signals`).

jaor:
Hmm. This is not working for me when I start more than one xmobar (I
usually have two running).  Sometimes the second one just exits, and the
other gives me the message "xmobar: too many pending signals" shortly
thereafter and dies... it's as if we were registering signals per
process name, rather than by pid, if that makes any sense...

[After trying a few more times]

It seems to be systematic: the second xmobar that I start eventually
dies with the too many pending signals message...

DanielSchuessler:
Weird. Is something sending tons of signals to your xmobar for some reason? The handler I added does `putMVar`, so it blocks inside the signal handler for all signals after the first one, leading to pending signals (should probably better be a `tryPutMVar`).

Killing individual xmobars works for me:
```
> xmobar &
[1] 5322
> xmobar &
[2] 5332
> pgrep xmobar
5322
5332
> kill 5322
> bar: Caught signal 15; exiting...
> pgrep xmobar
5332

```

jaor:
On Fri, Aug 31 2018, DanielSchuessler wrote:

> Weird. Is something sending tons of signals to your xmobar for some
> reason? The handler I added does putMVar, so it blocks inside the
> signal handler for all signals after the first one, leading to pending
> signals (should probably better be a tryPutMVar).

with tryPutMVar the second instance takes a bit longer to die, but it
eventually dies... i'm not aware of any signal being sent to it, no...

killing individual instances works for me too, that's not the problem.

but it's very systematic and happens with different xmobar configs, even
without Alsa running...

jaor:
hi @DanielSchuessler, sorry for the long hiatus.  I'm playing with this again, using the latest released xmobar and merging on top (there's only a couple of easy-to-solve conflicts), and also xmonad instead of sawfish.  the good news is that i don't see the problems i was describing with too many interruptions (so i am guessing that's something specific to sawfish)... if i don't see any other crash, i'd like to merge into master: let me know if you prefer to do the rebase yourself (there is a silly compilation warning that could be fixed too), or you'd rather make me push mine.  Thanks!

DanielSchuessler:
Rebased it to squash some of the minor commits.

About the crashes: Maybe you could try `strace -e signal xmobar` to see which signals are responsible?

TBH, I'm not certain about the full implications of catching all these exotic signals (maybe we should get a second opinion?). OTOH, I only catch them to get a chance to do cleanup, and re-raise the signal in the finally block after restoring the default handler.

Just ignoring the dangling `alsactl` processes would also be an option - as mentioned, the next volume change seems to kill them due to broken pipe, at least on my system.

jaor:
btw, i still see that couple of warnings:

```
src/Plugins/Monitors/Volume.hs:190:17: warning: [-Wname-shadowing]
    This binding for ‘opts’ shadows the existing binding
      bound at src/Plugins/Monitors/Volume.hs:135:15
    |
190 |     getFormatDB opts (Just d) = formatDb opts d
    |                 ^^^^

src/Plugins/Monitors/Volume.hs:194:21: warning: [-Wname-shadowing]
    This binding for ‘opts’ shadows the existing binding
      bound at src/Plugins/Monitors/Volume.hs:135:15
    |
194 |     getFormatSwitch opts (Just sw) = formatSwitch opts sw
    |                     ^^^^
```

DanielSchuessler:
Fixed. Unrelated to ALSA, but I also get these:
```
<no location info>: warning: [-Wmissing-home-modules]
    These modules are needed for compilation but not listed in your .cabal file's other-modules: Plugins.DateZone
                                                                                                 Plugins.MBox
                                                                                                 Plugins.Mail
```

using these flags:

```
flags:
    xmobar:
        with_xft: true
        with_utf8: true
        with_alsa: true
        with_weather: true
```

Looks like the modules are still getting pulled into the compilation even with the plugins disabled?

jaor:
On Sat, Oct 06 2018, DanielSchuessler wrote:

> Rebased it to squash some of the minor commits.
>
> About the crashes: Maybe you could try strace -e signal xmobar to see which signals are responsible?

hmmm.  that produces the following trace over and over:

rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
rt_sigprocmask(SIG_BLOCK, [HUP INT QUIT ILL TRAP ABRT BUS FPE SEGV ALRM TERM XCPU XFSZ PROF IO SYS], [], 8) = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
rt_sigprocmask(SIG_BLOCK, [HUP INT QUIT ILL TRAP ABRT BUS FPE SEGV ALRM TERM XCPU XFSZ PROF IO SYS], [], 8) = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0


at a rate of a few times per second... perhaps we are (re)setting the
mask each refresh?  what do you see in your system?

jaor:
On Sat, Oct 06 2018, DanielSchuessler wrote:

> Looks like the modules are still getting pulled into the compilation
> even with the plugins disabled?

Yes, Config.hs seems to be importing (and using) them unconditionally.
I'll fix it.

jaor:
another thing i am seeing is that, if i restart xmobar several times,
once in a while the processes don't get really killed and the volume
display doesn't get updated in the new instance, so maybe we're having
some kind of race condition with the signal handlers, or are violating
the OS protocol somehow... perhaps ignoring the processes is not that
bad a deal after all...

DanielSchuessler:
I see the `rt_sigprocmask` stuff too, but I assume it's just the GHC runtime blocking signals while it's doing something important. I get it on the master branch too, just with fewer signals: 

```rt_sigprocmask(SIG_BLOCK, [INT USR1 USR2], [], 8) = 0``` 

Just thought the `strace` might be useful to see what signal it caught before it crashed. You could also try commenting out all the handled signals except, say `sigTERM` and see if it still crashes (sorry, I can't reproduce it). If not, then one of the "weird" signals is probably responsible.

DanielSchuessler:
I suppose just `sigTERM` and `sigQUIT` (`sigINT` is already handled by GHC) would also cover the major use cases. Who kills their processes with `sigXFSZ`? ;)

jaor:
other than that, i'm not seeing crashes so far, not even with sawfish...

jaor:
On Sat, Oct 06 2018, DanielSchuessler wrote:

> I suppose just sigTERM and sigQUIT (sigINT is already handled by GHC)
> would also cover the major use cases. Who kills their processes with
> sigXFSZ? ;)

yes, that's true.  and if anybody does, they deserve a dangling process
;)

maybe we could just catch those ones and call it a day.

DanielSchuessler:
Sounds good :)

jaor:
okay, everything looks good, and i haven't had a crash in a good while, so let's give this a try :) merged, hanks for your patience!

PR #366, opened 2018-10-27
Return a dbus error if unknown method or signal

mitchellwrosen:
This patch gives a better error when you've tried calling a non-existent method, or a signal type that doesn't parse

### DIFF
--- a/src/IPC/DBus.hs
+++ b/src/IPC/DBus.hs
@@ -17,7 +17,7 @@ module IPC.DBus (runIPC) where
 import DBus
 import DBus.Client hiding (interfaceName)
 import qualified DBus.Client as DC
-import Control.Monad (when)
+import Data.Maybe (isNothing)
 import Control.Concurrent.STM
 import Control.Exception (handle)
 import System.IO (stderr, hPutStrLn)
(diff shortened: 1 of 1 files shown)


jaor:
@mitchellwrosen thanks, this doesn't look bad... but it's producing hlint warnings (see the build reports): could you please take a look at them?

mitchellwrosen:
Done!

mitchellwrosen:
I ended up here while trying to implement a feature - update my list of active WireGuard interfaces when they change (rather than on a timer).

I didn't need this code in the end, it just helped debug what was going on with dbus. I've discovered that xmobar plugins can't react to signals directly, they are just `start`ed and run forever.

So instead I'm using a fifo and a `PipeReader`. Works great!

jaor:
thanks @mitchellwrosen ... i was looking at the code again, and, unless i am reading it wrong, we're changing the return value of that function when everything goes well, deleting information on which signals were processed. so i am not sure if this is granted, specially if you're not even using it.  thoughts?

mitchellwrosen:
@jaor What do you mean by this?

> we're changing the return value of that function when everything goes well, deleting information on which signals were processed

As far as I can tell, the behavior _before_ was:

- If the method name is "SendSignal", 
  - Attempt to parse and send signals
  - Return OK
- Else return OK

Whereas now it's

- If the method name is "SendSignal",
  - Attempt to parse and send signals
  - If _any_ signal failed to parse,
    - Return `errorInvalidParameters` (even if _some_ signal did successfully parse)
  - Else return OK
- Else return `errorUnknownMethod`

Or am I mistaken about something?

jaor:
oh, sorry mitchell... it just means that i misread our previous
implementation! merged, thanks again!

mitchellwrosen:
No worries at all, thanks!

PR #367, opened 2018-10-30
Allow recovery from missing thermal zones

bobtwinkles:
Sometimes the kernel takes some time to populate /sys/class/thermal/ * after
resuming from suspend. Previously, this would result in the thermal plugin
dying permanantly. With this patch, we show "0" until the kernel gets around to
populating the sysfs nodes we need.

There's something to be said for generalizing this fix and applying it to the
other monitor plugins that rely on sysfs nodes. ThermalZone is the only one
I rely on personally so that's the one that I fixed :wink:.


bobtwinkles:
I felt bad about how hacky my first pass at this was, so I've updated this PR with slightly less hacky handling. We now display "updating..." while waiting for the thermal zones to come back up.

jaor:
merged, thanks.  i agree we should generalize this idea to other monitors...

PR #377, opened 2019-01-19
Alsa plugin: Restart alsactl if it quits (Fixes #376)

### LINKED ISSUE #376: Alsa plugin errors after suspend
The Alsa plugin produces an error after my computer suspends.
The error is displayed in the status bar: `error: ExceptionInLinkedThread ThreadId 24 fd:11: hGetChar: end of file`

Here's the Alsa portion of my config:
```
, Run Alsa "default" "Master"
	[ "--template", "Vol: <volume>% <status>"
	, "--Low",			"33"			-- units: %
	, "--High",		"66"			-- units: %
	, "--low",			"darkred"
	, "--normal",		"darkorange"
	, "--high",			"darkgreen"
	]
```

While I don't know what's causing the issue, I was wondering if there is a way to clear the error from the status bar?

DanielSchuessler:
See also: #376

Is it safe to concurrently call the plugin output callback? Otherwise, I'd have to amend this because a concurrent write is theoretically possible if the main plugin thread is lagging and still processing an event from before `alsactl` crashed, while the reader thread writes "Restarting alsactl...".

I think it is safe because apparently, the callback ultimately comes from here:

```haskell
startCommand :: [...]
startCommand sig (com,s,ss)
[...]
    | otherwise = do var <- atomically $ newTVar is
                     let cb str = atomically $ writeTVar var (s ++ str ++ ss)
                     a1 <- async $ start com cb
                     a2 <- async $ trigger com $ maybe (return ())
                                                 (atomically . putTMVar sig)
                     return ([a1, a2], var)
    where is = s ++ "Updating..." ++ ss

```

### DIFF
--- a/src/Xmobar/Plugins/Monitors/Alsa.hs
+++ b/src/Xmobar/Plugins/Monitors/Alsa.hs
@@ -87,12 +87,12 @@ startAlsaPlugin mixerName controlName args cb = do
     runMB args Volume.volumeConfig run wait_ cb
 
 withMonitorWaiter :: String -> Maybe FilePath -> (String -> IO ()) -> (IO () -> IO a) -> IO a
-withMonitorWaiter mixerName alsaCtlPath outputCallback cont = do
+withMonitorWaiter mixerName alsaCtlPathOverride outputCallback cont = do
   mvar <- newMVar ()
 
-  path <- determineAlsaCtlPath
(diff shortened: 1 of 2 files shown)


jaor:
@DanielSchuessler thanks for the fix.  Yes, since we're using TVars, concurrent calls to the callback should be atomic.

PR #383, opened 2019-03-01
Add `nix` section to stack.yaml

avdv:
When building xmobar on NixOS (where stack's Nix integration is enabled by
default), this ensures that all required system libraries are available inside
the local build environment.

See https://docs.haskellstack.org/en/stable/nix_integration/ for details.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -18,3 +18,16 @@ extra-deps:
 # - parsec-numbers-0.1.0
 # - libmpd-0.9.0.9
 # - alsa-mixer-0.3.0
+
+nix:
+    packages:
+      - alsaLib
+      - pkgconfig
+      - wirelesstools
+      - xorg.libX11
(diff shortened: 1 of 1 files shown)


jaor:
@avdv is there any relationship with this pr and #384? sorry but i've never used nix, and both prs are claiming to support it in ways that look very different..

avdv:
These PRs actually do different things. #384 adds a Nix "expression" which allows anyone to build a proper nix package from the sources. Usually, AFAIUI, those expressions are added to some nix repository like the official https://github.com/NixOS/nixpkgs channels and not to a project itself, but I may be wrong.

This PR allows you to checkout the sources and just build the project with the stack tool, which takes care that the required dependencies are available in the build environment were ghc / cabal is running.

jaor:
thanks for the clarification, merged!

PR #384, opened 2019-03-01
Added support for building with Nix and NixOS

BadDecisionsAlex:


jaor:
is there any relationship with this pr and #383? sorry but i've never
used nix, and both prs are claiming to support it in ways that look very
different..

jaor:
rebased and merged, thanks!

BadDecisionsAlex:
Hey sorry for the late response. This has some overlap with the other one that you mentioned, but they accomplish different things. Ideally both would be included. 
Shell.nix is a developer container for hacking around, while default.nix is the package installation script.

PR #394, opened 2019-07-10
Touched up CoreTemp for better readability

jumper149:
I want to implement a Plugin, which calculates the average core temperature.
I started with reading through the old CoreTemp Plugin and got to the conclusion, that I cannot calculate the average temperature, when the files are read by `checkedDataRetrieval`.

Along the way I made everything clear in the Plugin.

I would like your opinion on how to calculate an average temperature.
Is it necessary to manually read the files (like in the Cpu Plugin)?


jumper149:
Just as I suggested I rewrote the whole plugin in the style of the Cpu plugin.
I will make a new Pull Request for that, as I think this plugin should be kept for backwards-compatability in configs.

PR #400, opened 2019-10-10
Fix type errors in MPD plugin

johntyree:
I don't have an MPD install to test with right now, but right now you can't bulid xmobar with `all_extensions: true` at all.

### DIFF
--- a/src/Xmobar/Plugins/Monitors/MPD.hs
+++ b/src/Xmobar/Plugins/Monitors/MPD.hs
@@ -99,13 +99,13 @@ parseMPD (Right st) song opts = do
         si = stateGlyph s opts
         vol = int2str $ fromMaybe 0 (M.stVolume st)
         (p, t) = fromMaybe (0, 0) (M.stTime st)
-        [lap, len, remain] = map showTime [floor p, t, max 0 (t - floor p)]
-        b = if t > 0 then realToFrac $ p / fromIntegral t else 0
+        [lap, len, remain] = map showTime [floor p, floor t, max 0 (floor t - floor p)]
+        b = if t > 0 then realToFrac $ p / t else 0
         plen = int2str $ M.stPlaylistLength st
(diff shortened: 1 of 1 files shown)


jaor:
would you mind showing me the errors you get? i build xmobar almost
daily several times (i compile it from its configuration), always with
all_extensions set to true, without any trouble, so i'm very puzzled by
what you say...

src/Xmobar/Plugins/Monitors/MPD.hs
```diff
@@ -99,13 +99,13 @@ parseMPD (Right st) song opts = do
...
-        [lap, len, remain] = map showTime [floor p, t, max 0 (t - floor p)]
-        b = if t > 0 then realToFrac $ p / fromIntegral t else 0
+        [lap, len, remain] = map showTime [floor p, floor t, max 0 (floor t - floor p)]
+        b = if t > 0 then realToFrac $ p / t else 0
         plen = int2str $ M.stPlaylistLength st
         ppos = maybe "" (int2str . (+1)) $ M.stSongPos st
         flags = playbackMode st
 
-stateGlyph :: M.State -> MOpts -> String
+stateGlyph :: M.PlaybackState -> MOpts -> String
```
jaor: oh, i see now, this new type must have been introduced in libmpd 0.9.1, because this code is compiling fine by me with 0.9.0 (i'm a bit surprised a type change like this happened on a micro-version update, maybe i am missing something here).  would you mind also changing the required libmpd version in xmobar.cabal? thanks a lot!

jaor:
never mind @johntyree, i've bumped the version in the cabal file and merged your commit together with it. thanks again!

src/Xmobar/Plugins/Monitors/MPD.hs
johntyree: I agree, that's a big change for a minor version bump. Maybe they didn't realize how disruptive it would be? I'd call it out upstream.

PR #401, opened 2019-10-10
Update state names and value conversion due to libmpd-haskell updates

akilegaspi:
Some updates on the libmpd-haskell package has been done and compiling from source does yields an error due to it.

This commit fixes these compilation errors.


akilegaspi:
nvm, there was already a proposed fix for this

PR #403, opened 2019-10-12
Display additional string depending on battery status

slotThe:
This extends the `Battery` plugin by three new arguments: `--lows`, `--mediums`, and `--highs` (I could not think of any mnemonically sound one character aliases for these). They can be used to display an additional string, depending on the current battery level (as set by `-L` and `-H`).  It be displayed in front of the `--off` string. All three values default to `""`, so as to not interfere with any configuration not explicitly setting them.

Usage example:
```
Config { font = "xft:FontAwesome:antialias=true:hinting=true:size=9,sans"
       , commands = [ Run Battery
                          [ "--template", "<acstatus>"
                          , "--"
                          , "--lowt"    , "15"
                          , "--hight"   , "70"
                          , "--lows"    , ""
                          , "--mediums" , ""
                          , "--highs"   , ""
                          , "--off"     , "  <left>% (<timeleft>)"
                          ] 100
                    ]
       , template = "%battery%"
       }
```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,8 @@ _New features_
   - New option `--contiguous-icons` for `MultiCpu` to draw icons
     without padding (see [issue #388]).
   - New version of libmpd (0.9.0.10), thanks to John Tyree
+  - New options `--lows`, `--mediums`, and `--highs` for `Battery`
+    to display an additional string depending on battery level.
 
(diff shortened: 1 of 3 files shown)


jaor:
thanks, this looks good.  i've merged it, fixed the changelog (0.31 is
already released) and replaced the use of `<>` to make GHC 8.0 and 8.2
happy.

PR #406, opened 2019-10-14
Display additional string based on volume level

slotThe:
This basically implements the functionality from #403 for the `Volume` plugin. It adds five new arguments: `-L (--lowv)`, `-H (--highv)`, `-l (--lows)`, `-m (--mediums)`, and `-h (--highs)`. 
`-L` and `-H` are used to set percentage-based thresholds for the volume level. These two values default to `Just 20.0` and `Just 60.0` respectively. 
`-l`, `-m`, and `-h` can then be used to display an additional string, depending on the current volume level. This will only be displayed in front of the string defined by `-O`. All three values default to "", so as to not interfere with any configuration not explicitly setting them.

As for the commits:
  1. Adds the types for `-L` and `-H` to `VolumeOpts`, sets some defaults and adds `Option`s
  2. Does the same as 1. for `-l`, `-m`, and `-h`
  3. Adds the `VolumeStatus` type for convenient pattern matching and ways to convert between a volume percentage and a `VolumeStatus`, as well as a `VolumeStatus` and a string as defined by `-l`, `-m`, or `-h`.
  4. Changes the volume formatting so as to print the above defined strings in front of the `onString`.
  5. Updates the readme.
  6. Updates the changelog.

Usage example:
```
Config { font = "xft:FontAwesome:antialias=true:hinting=true:size=9,sans"
       , commands =  [Run Volume "default" "Master"
                       [ "--template", "<status> <volume>%"
                       , "--"
                       , "--on"     , ""
                       , "--lows"   , ""
                       , "--mediums", ""
                       , "--highs"  , ""
                       ] 10
       ]
       , template = "%default:Master%"
       }
```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,9 @@ _New features_
 
   - New options `--lows`, `--mediums`, and `--highs` for `Battery`
     to display an additional string depending on battery level.
+  - New options `-L` and `-H` for `Volume` to set low and high volume
+    levels, as well as `-l`, `-m`, and `-h` to display an additional
+    string depending on current volume level.
 
 
 ## Version 0.31 (October, 2019)
--- a/readme.md
+++ b/readme.md
@@ -1155,6 +1155,25 @@ more than one battery.
     - `--highd` _number_ High threshold for dB. Defaults to -5.0.
     - `--lowd` _number_ Low threshold for dB. Defaults to -30.0.
     - `--volume-icon-pattern` _string_ dynamic string for current volume in `volumeipat`.
+    - `-H` _number_ High threshold for volume (in %). Defaults to 60.0.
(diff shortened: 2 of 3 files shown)


jaor:
merged, thanks! (unless you prefer otherwise, i'd like to mention you in the readme/changelog as the contributor of these improvements, but the only id i have is slotThe. should i use that? it's perfectly okay if you rather not, just ignore this note if so!)

slotThe:
Thanks for the merge!

As of right now I wouldn't feel entirely confident giving out my real name.  If you want to thank me under "slotThe", however, I would appreciate it :)

jaor:
thanks duly added to changelog.md :)

PR #407, opened 2019-10-14
Allow latest GHC

vmchale:
Bump to allow GHC 8.8.1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ ghc:
   - 8.2
   - 8.4
   - 8.6
+  - 8.8
 
 before_install:
   - sudo apt-get -qq update
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ before_install:
   - export PATH=/opt/ghc/bin:$PATH
 
 install:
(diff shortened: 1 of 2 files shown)


jaor:
thanks, @vmchale ... from the travis build, it seems we have some other dependency to be updated, but i am not sure which one: https://travis-ci.org/jaor/xmobar/jobs/597856087 . any ideas? (sorry, right now i don't have a 8.8 installation handy).

vmchale:
It seems to be something related to `cabal-install` 2.0 unfortunately. It works fine with `cabal-install` 2.2. 

Either we could disable the CI for 8.8, or it could be fixed to use a more modern `cabal-install`.

jaor:
> Either we could disable the CI for 8.8, or it could be fixed to use a
> more modern cabal-install.

do you know if the latter would be complex? i'd prefer not disabling it,
but if fixing it gives us too much trouble we might as well wait until
travis uses it.

vmchale:
I'll try to use cabal-install 2.2 in CI; it seems to work now?

jaor:
excellent! thanks a lot for taking care of this.  rebased and merged!

PR #408, opened 2019-10-20
add filtering option for Net devices

vindex10:
Add option to specify which network devices to show. The option is relevant for instance when docker network interfaces are not of interest:

```
× sit0@NONE
√ wlp3s42
√ enp0s20f41
× br-uh14b3h3v
× docker0
× ethekjbrh14@if7
```
Example:
```
DynNetwork ["--", "--devices", "wlp3s42,enp0s20f41"] 10
```

### DIFF
--- a/src/Xmobar/Plugins/Monitors/Common/Output.hs
+++ b/src/Xmobar/Plugins/Monitors/Common/Output.hs
@@ -17,8 +17,6 @@
 
 module Xmobar.Plugins.Monitors.Common.Output ( IconPattern
                                              , parseIconPattern
-                                             , DevList
-                                             , parseDevList
                                              , padString
                                              , showWithPadding
                                              , showWithColors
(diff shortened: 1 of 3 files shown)


src/Xmobar/Plugins/Monitors/Common/Output.hs
```diff
@@ -17,6 +17,8 @@
 
 module Xmobar.Plugins.Monitors.Common.Output ( IconPattern
                                              , parseIconPattern
+                                             , DevList
+                                             , parseDevList
```
jaor: this is a minor thing, but i think that, given that parseDevList is used only in Net.hs, i would define it there (as well as DevList) and not even export it.

jaor:
another minor thing, but perhaps we could call the switch `--devices`, which is shorter and easier to remember?

jaor:
oh, and would you add a mention of the new option to the corresponding section in the readme? thanks a lot!

jaor:
squashed and merged, thanks a lot!

vindex10:
my pleasure!)

PR #409, opened 2019-10-28
Optimize Net a bit (noticeably less CPU/power usage)

liskin:
- use strict ByteString as the Lazy version of readFile allocates a 32k
  buffer even though we usually need much less (isUp needs a few bytes)

- refactor NetDev datatype and use unsafeInterleaveIO in isUp to avoid
  reading the operstate file entirely if we're not interested in that
  device

- postpone ByteString unpacking in netParser to shave off some cycles,
  and avoid ByteString unpacking in isUp entirely

On my system with 8 network devices (and more if docker is up), this
seems to reduce xmobar's CPU usage noticeably. I have two "Run Network"
in xmobar configuration, for eth and wlan, so without these changes,
xmobar would evaluate isUp 16 times a second, and each evaluation would
allocate a buffer for the IO Handle and then another buffer for the lazy
ByteString readFile. Now it only does isUp once for every device I'm
interested in, and the only large buffers allocated are the IO Handle
ones (getting rid of these isn't worth the code complexity).

### DIFF
--- a/src/Xmobar/Plugins/Monitors/Net.hs
+++ b/src/Xmobar/Plugins/Monitors/Net.hs
@@ -13,6 +13,8 @@
 --
 -----------------------------------------------------------------------------
 
+{-# LANGUAGE OverloadedStrings #-}
+
 module Xmobar.Plugins.Monitors.Net (
                         startNet
                       , startDynNet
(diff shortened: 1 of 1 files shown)


jaor:
looks good, merged! thanks, and sorry for the delay

liskin:
No worries, 9 days is no delay. :-))

PR #410, opened 2019-10-29
Implement timer coalescing (noticeably less CPU/power usage)

liskin:
xmobar currently runs every monitor in its own thread. Monitors that do periodic updates simply sleep and loop. This unfortunately leads to these threads coming out of sync, and xmobar ends up waking up and redrawing for every periodic monitor. In my case, that is 7 times per second, which is enough for xmobar to be at the top of "top" with more than 1% CPU usage, and to have a noticeable impact on battery life.

This commit adds a central timer coordination thread which makes sure that periodic updates happen together and that we only redraw once they're all done.

Together with PR #409, I managed to lower the idle power draw of my laptop from 4W to 3W.

### DIFF
--- a/src/Xmobar/App/EventLoop.hs
+++ b/src/Xmobar/App/EventLoop.hs
@@ -47,7 +47,10 @@ import Xmobar.X11.Text
 import Xmobar.X11.Draw
 import Xmobar.X11.Bitmap as Bitmap
 import Xmobar.X11.Types
+
+#ifndef THREADED_RUNTIME
 import Xmobar.X11.Events(nextEvent')
+#endif
 
 #ifdef XFT
 import Graphics.X11.Xft
--- a/src/Xmobar/App/EventLoop.hs
+++ b/src/Xmobar/App/EventLoop.hs
@@ -17,7 +17,12 @@
 ------------------------------------------------------------------------------
 
 
-module Xmobar.App.EventLoop (startLoop, startCommand) where
+module Xmobar.App.EventLoop
+    ( startLoop
+    , startCommand
+    , newRefreshLock
+    , refreshLock
+    ) where
 
 import Prelude hiding (lookup)
 import Graphics.X11.Xlib hiding (textExtents, textWidth)
@@ -31,7 +36,7 @@ import Control.Monad.Reader
 import Control.Concurrent
 import Control.Concurrent.Async (Async, async)
 import Control.Concurrent.STM
-import Control.Exception (handle, SomeException(..))
+import Control.Exception (bracket_, handle, SomeException(..))
 import Data.Bits
 import Data.Map hiding (foldr, map, filter)
 import Data.Maybe (fromJust, isJust)
@@ -63,15 +68,34 @@ import Xmobar.System.DBus
 runX :: XConf -> X () -> IO ()
 runX xc f = runReaderT f xc
 
+newRefreshLock :: IO (TMVar ())
+newRefreshLock = atomically $ newTMVar ()
+
+refreshLock :: TMVar () -> IO a -> IO a
+refreshLock var = bracket_ lock unlock
+    where
+        lock = atomically $ takeTMVar var
+        unlock = atomically $ putTMVar var ()
+
+refreshLockT :: TMVar () -> STM a -> STM a
+refreshLockT var action = do
+    takeTMVar var
+    r <- action
+    putTMVar var ()
+    return r
+
 -- | Starts the main event loop and threads
-startLoop :: XConf -> TMVar SignalType -> [[([Async ()], TVar String)]]
-             -> IO ()
-startLoop xcfg@(XConf _ _ w _ _ _ _) sig vs = do
+startLoop :: XConf
+          -> TMVar SignalType
+          -> TMVar ()
+          -> [[([Async ()], TVar String)]]
+          -> IO ()
+startLoop xcfg@(XConf _ _ w _ _ _ _) sig pauser vs = do
 #ifdef XFT
     xftInitFtLibrary
 #endif
     tv <- atomically $ newTVar []
-    _ <- forkIO (handle (handler "checker") (checker tv [] vs sig))
+    _ <- forkIO (handle (handler "checker") (checker tv [] vs sig pauser))
 #ifdef THREADED_RUNTIME
     _ <- forkOS (handle (handler "eventer") (eventer sig))
 #else
@@ -111,15 +135,16 @@ checker :: TVar [String]
            -> [String]
            -> [[([Async ()], TVar String)]]
            -> TMVar SignalType
+           -> TMVar ()
            -> IO ()
-checker tvar ov vs signal = do
-      nval <- atomically $ do
+checker tvar ov vs signal pauser = do
+      nval <- atomically $ refreshLockT pauser $ do
               nv <- mapM concatV vs
               guard (nv /= ov)
               writeTVar tvar nv
               return nv
       atomically $ putTMVar signal Wakeup
-      checker tvar nval vs signal
+      checker tvar nval vs signal pauser
     where
       concatV = fmap concat . mapM (readTVar . snd)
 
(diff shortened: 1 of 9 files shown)


jaor:
hmm, this looks very interesting, but i'm not sure we don't need some additional tweak.  for instance, i use a weather monitor that has a very long period (30 minutes), and what i'm observing playing with this patch is that all my monitors are falling back to wait 30 minutes displaying "Updating...".  we cannot simply wait to update everything at the end (also, people might in fact want different rates, even without such a big difference): am i misunderstanding or is this what these changes implement? (if the later, we could still retain them for your use case, but with an opt-in flag in the configuration to enable this behaviour)...

liskin:
@jaor That seems to be a bug. My configuration mixes once per second and once per 5 seconds updates and it works well. Date/time is updated once a second and batteries are queried/updated once every 5 seconds. Also, I think I made sure that all monitors update at the beginning.

But now that I think about it, there is a bug indeed: if a monitor takes a long time to refresh, then all other refreshes are delayed. So I'll probably need to only apply the lock to the callback that pushed the updates to the display loop thread.

liskin:
Or maybe just add a timeout or some way to designate that a monitor is expensive and that we shouldn't postpone the refresh. I'll think about it some more. Didn't realize there were monitors like this as I only ever used those more frequent ones.

jaor:
ah, that makes sense. skimming over the code i got the impression that
you were already taking care of uneven periods, but running it produced
that behaviour.

i can actually confirm that removing the weather monitor in my config
with that long period seems to be fixing the rest!

jaor:
ideally, an initial, unconditional refresh of all monitors at t=0 would
fix the problem, i guess...

liskin:
The `newPeriod` function sets `next` to something that's `≤ now`, so there _is_ an unconditional refresh of all monitors. Are you sure it's not just taking too long?

Anyway, I'll try the weatcher monitor myself later. And I think I saw your configuration somewhere on github, so I can even test with your config. :-)

jaor:
On Wed, Nov 06 2019, Tomáš Janoušek wrote:

> The newPeriod function sets next to something that's ≤ now, so there
> is an unconditional refresh of all monitors. Are you sure it's not
> just taking too long?

Hmm, i'll doble check, but starting the monitor with the previous
version of xmobar shows the weather in a matter of secs, and i had all
my monitors showing "Updating..." for something like 10 minutes with
this one.

> Anyway, I'll try the weatcher monitor myself later. And I think I saw
> your configuration somewhere on github, so I can even test with your
> config. :-)

Actually, i forgot to mention that these days i use xmobar as a library
and compile my configuration (maybe that's having an effect here):

  https://gitlab.com/jaor/xmobar-config

stack install in that repo, with an xmobar checkout at the same level as
xmobar-config, yields three executables... i'm using xmobar-exwm.
here's the weather "config":

  https://gitlab.com/jaor/xmobar-config/blob/master/src/lib/Monitors.hs#L79

(`p` there is just containing some colors that switch automatically
between dark and light themes from an env var, so you can probably
ignore that).

jaor:
er, i just checked again and seems to be working fine also with a
weather monitor, so you must be right and it was just the weather
station not answering on time... i'm a bit worried that a single monitor
can block the rest like that, but i'm more worried about battery
consumption :)

liskin:
Okay, thanks for testing and confirming my hypothesis! I'll try to come up with a solution for that blocking problem, I think it's serious enough.

liskin:
It would block the entire xmobar whenever you lose network connection. That's not acceptable, even if it saved 10 watts. :-)

jaor:
On Wed, Nov 06 2019, Tomáš Janoušek wrote:

> It would block the entire xmobar whenever you lose network
> connection. That's not acceptable, even if it saved 10 watts. :-)

you've got a point! :)

liskin:
@jaor Hi! Sorry for the delay, I was busy with moving to another country and work and other stuff.

This weekend I rebased on top of master and tried to fix the slow monitor issue. Took me a while to figure out how to make the slow monitor detection work without introducing another wakeup for the timeout, so I ended up using ```next timer `max` one second``` which should be fine power-consumption-wise and it should also feel reasonably interactive. I just finished this a few minutes ago so it's somewhat untested. Feel free to test it, but no promises! :-)

And no hurries, too. Another work week awaits. :-)

jaor:
hi! glad to see you're still around :) i like the general idea behind
these changes a lot.

that said, i'm afraid your last changes didn't work.  i just compiled
and ran, and all i'm seeing in my xmobar for *all* my monitors is
"Updating..." (i'm guessing that due to a weather with 30min retry
time).

i haven't had time to look at the code yet, so i'm afraid i cannot
give you more useful feedback.  the behaviour seems deterministic.

liskin:
Oh, there's one other thing I forgot. `registerDelay` only works with the threaded RTS so I need to disable the whole thing if `with_threaded` cabal flag is off (which I think it might be in [your case](https://gitlab.com/jaor/xmobar-config/-/blob/master/stack.yaml). And then document it somewhere...

liskin:
(just a force-push to fix a hlint warning)

Anyway, I just recompiled without threaded and I'm seeing the same behaviour as you do. It's a bit unfortunately that there's no mention of the `errorWithoutStackTrace "registerDelay: requires -threaded"` exception in ~/.xsession-errors. :-(
if you get the chance to try `with_threaded`, that would be awesome. I'm afraid this really won't work without it, ever.

jaor:
indeed, things are working much better now for me, i'll keep you posted!

i'm wondering whether we should make threaded the default, the problems
in issue #36 are 8 years old...

jaor:
> if you get the chance to try with_threaded, that would be awesome. I'm
> afraid this really won't work without it, ever.

yes, i understand that. one more reason to make threaded the default...

liskin:
I found https://hackage.haskell.org/package/stm-delay, so apparently it is possible to do this the non-threaded RTS somehow. Whether it's worth it, I don't know. Unfortunately I'm no longer using Haskell daily so I'm totally out of touch with the current state of affairs regarding threaded RTS trade-offs. :-/

jaor:
i think i would go for the option of disabling coalescing in the
non-threaded case, if that's not too cumbersome.  (unfortunate indeed
that you are not using haskell daily! but you're not alone :))

liskin:
@jaor Implemented disabling of coalescing in the non-threaded case by implementing proper robust exception handling. More details in the commit message. Hope it's all right now. :-)

jaor:
hmm, i'm getting a compilation error when with_threaded is false:

```
/home/jao/usr/jao/xmobar/xmobar/src/Xmobar/App/Timer.hs:187:5: error:
    • No instance for (Control.Monad.Fail.MonadFail STM)
        arising from a do statement
        with the failable pattern ‘Just periods’
    • In a stmt of a 'do' block: Just periods <- readTVar periodsVar
      In the second argument of ‘($)’, namely
        ‘do Just periods <- readTVar periodsVar
            periods' <- foldrM unCoalesceTimer periods timers
            writeTVar periodsVar $ Just periods'’
      In the expression:
        atomically
          $ do Just periods <- readTVar periodsVar
               periods' <- foldrM unCoalesceTimer periods timers
               writeTVar periodsVar $ Just periods'
    |
187 |     Just periods <- readTVar periodsVar
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

liskin:
Oh, darn. Travis gets that error as well. I wonder why I don't... :-/

liskin:
Should be fixed now but I'm still very confused why it failed in Travis as well. MonadFailDesugaring was only enabled in GHC 8.6 or something and the warning was never enabled by default. How on earth is that an error in GHC 8.0.2 on Travis?!

jaor:
beats me too... i've merged the pr nonetheless! it's working fine for
me.  i've left the default for with_threaded to false for now, but
updated the stack file with it set to true, as a hint ;)

thanks a lot!

liskin:
Thanks for your quick feedback!

PR #411, opened 2019-11-17
DateZone: get timezone series from TZDIR

emmanuelrosa:
The DateZone plugin calls `getTimeZoneSeriesFromOlsonFile` using the
hard-coded path /usr/share/zoneinfo. While that may work just fine on
most Linux distros, it does not work on NixOS since that directory is
always located somewhere under /nix/store.

Based on mild research, it seems the environment variable TZDIR is
commonly set to the absolute path to `zoneinfo` (but without a trailing
slash).

This change modifies the DateZone plugin to first try getting
the zoneinfo path from the TZDIR environment variable, falling back
to the hard-coded path /usr/share/zoneinfo


jaor:
thanks!

PR #414, opened 2019-12-23
Add support for the "Weather" field

### LINKED ISSUE #254: Weather plugin -- parse and exponse the "Weather:" line of the data
Consider the following data (taken from current measurements at the "LIBP" station):

``` text
Bolzano, Italy (LIPB) 46-28N 011-20E 1461M
Jan 11, 2016 - 09:50 AM EST / 2016.01.11 1450 UTC
Wind: Variable at 2 MPH (2 KT):0
Visibility: less than 1 mile:0
Sky conditions: overcast
Weather: light rain; mist
Temperature: 42 F (6 C)
Dew Point: 42 F (6 C)
Relative Humidity: 100%
Pressure (altimeter): 29.26 in. Hg (0991 hPa)
ob: LIPB 111450Z VRB02KT 1500 -RA BR SCT001 OVC020 06/06 Q0991
cycle: 15
```

The `Weather:` line tells us what the current weather actually is like; however, the xmobar plugin does not expose it to the user.  Is there any reason why this is so?

slotThe:
Closes #254 by adding the necessary functionality.

Now, this is by no means a perfect solution.  As you said in the original issue, optional fields are sort of a pain and I did not think this "important" enough to extend the `WeatherX` extension (thus most likely breaking existing configs).  Feel free to do this (or yell at me to do so) if you think it is worth it :)
I also don't know if @d125q's request was to simply spit this out as a string, or if he wanted any more processing done on this.  Alas, the issue is very old and so getting a response there seems rather hopeless.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,11 @@
-## Version 0.32
+## Version 0.33
+
+_New features_
+
+  - New template parameter `<weather>` for the `Weather` plugin, potentially
+    displaying specific weather conditions that are occurring near the station.
+
+## Version 0.32 (December, 2019)
 
 _New features_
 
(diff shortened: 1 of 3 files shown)


jaor:
this one looks good enough to me, at least for the time being. thanks!

PR #416, opened 2020-01-02
Add options to the Weather plugin

slotThe:
Add a `WeatherOpts` type (plus a first option) to the weather plugin.  

The original motivation was to build upon #414, this time perhaps handling the "fields doesn't exist" case a bit more gracefully.

The existence of yet another identical `parseOpts` function is not ideal, but this pr might not be the best place for cleaning this up (I plan on actually doing that in a separate pr).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -735,7 +735,11 @@ something like:
 - Aliases to the Station ID: so `Weather "LIPB" []` can be used in
   template as `%LIPB%`
 - Thresholds refer to temperature in the selected units
-- Args: default monitor arguments
+- Args: default monitor arguments, plus:
+  - `--weathers` _string_ : display a default string when the `weather`
+    monitor is not reported.
+    - short option: `-w`
+    - Default: ""
 - Variables that can be used with the `-t`/`--template` argument:
 	    `station`, `stationState`, `year`, `month`, `day`, `hour`,
 	    `windCardinal`, `windAzimuth`, `windMph`, `windKnots`, `windMs`, `windKmh`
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -735,7 +735,11 @@ something like:
 - Aliases to the Station ID: so `Weather "LIPB" []` can be used in
   template as `%LIPB%`
 - Thresholds refer to temperature in the selected units
-- Args: default monitor arguments
+- Args: default monitor arguments, plus:
+  - `--weathers` _string_ : display a default string when the `weather`
+    monitor is not reported.
+    - short option: `-w`
+    - Default: "normal"
```
jaor: hmm, i would perhaps make "" the default?

src/Xmobar/Plugins/Monitors/Weather.hs
```diff
@@ -26,6 +26,37 @@ import qualified Data.ByteString.Lazy.Char8 as B
...
+
+-- | Options the user may specify.
+data WeatherOpts = WeatherOpts
+  { weatherString :: String
+  }
+
+-- | Default values for options.
+defaultOpts :: WeatherOpts
+defaultOpts = WeatherOpts
+  { weatherString = "normal"
```
jaor: perhaps "" is better? i'm not sure what "normal" weather is :)

jaor:
thanks, this looks useful.  i commented on a little nit.  and yes, i
think it'll be good when we deal with `parseOpts`, but that can be
done separately.

slotThe:
Thanks for the review!

I agree that "normal" is perhaps a bit too opinionated as a default, so I have changed it to "" instead.

jaor:
merged, thanks!

PR #418, opened 2020-01-04
Replace parseOpts with a generic function

slotThe:
As promised in #416, this replaces the `parseOpts` function with a generic `parseOptsWith` one, in an effort to reduce code duplication.


jaor:
excellent, thanks a lot!

jaor:
(i've rebased and pushed your changes, closing manually)

PR #419, opened 2020-01-05
Add <volumestatus> template to Volume

### LINKED ISSUE #361: New option for Volume plugin
Can we get an extra option for template where it merges `<status>` and `<volume>` variables?

Something like, if it's off, just appears OFF (or another user defined message), if it's on, just appears the volume percentage.

slotThe:
Closes #361 by adding the provided functionality.  No new options were created,
as we already have quite usable `onString`, `offString`, as well as
`[low,mid,high]String` variables.  All get prepended to the volume with this, as
having `<status>` before `<volume>` seems most natural.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ tags
 /cabal.project.local
 /.ghc.environment.*
 /cabal.project.local~
+/stack.yaml.lock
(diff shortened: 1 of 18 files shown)


slotThe:
Whoops, looks like there are some commits in there that shouldn't be... 
the `Merge remote-tracking branch 'upstream/master'` seems very hairy to get out of there, if you want you can close this and I'll make a new pr with only the two relevant ones.

jaor:
On Sun, Jan 05 2020, slotThe wrote:

> if you want you can close this and I'll make a new pr with only the
> two relevant ones.

yes, please.  thanks!

Issue #415, opened 2020-01-02
[Advice needed] Battery Monitor plugin for FreeBSD


dbalan:
The `Battery` plugin doesn't work on FreeBSD. The plugin parses `sysfs` entries to get data which doesn't exist on FreeBSD. 

So a working `Battery` plugin on FreeBSD would have to query `sysctl` entries to work:

```
hw.acpi.battery.units: 1
hw.acpi.battery.state: 1
hw.acpi.battery.rate: 24076
hw.acpi.battery.time: 186
hw.acpi.battery.life: 90
```
If one were to add support for FreeBSD battery monitor, should this be a seperate plugin `FreeBSDBattery` or is there a way to integrate with current Battery plugin?

jaor:
there's no pre-defined way to integrate new batteries to the current
plugin, other than the fact that there's probably lots of code that can
be shared.  ideally, we would have just one plugin that would detect
automatically where it has to look for the info, but if that gets hairy
or makes the code more difficult to maintain or understand, there's
nothing wrong in a separate FreeBSDBattery (which could then be a
compilation option, i guess)...

dbalan:
Thanks @jaor I'll check if I can integrate the code into the current module if not I'll create a new one.

PR #421, opened 2020-01-06
Support for freebsd battery status:

dbalan:
Exposed via -f with_freebsd flag, uses sysctl to query battery status.

related to #415 

This is a work in progress, I tried making the current `Batt` plugin to work on both Linux and FreeBSD. On FreeBSD first argument to select `BAT` is ignored and the data queried is already aggregated.

The code is a bit hairy with preprocessor arguments in few places, I feel like I can refactor it to be bit nicer -- but I wanted to get a second opinion.

Thanks


jaor:
thanks! i don't think it's that bad, taking into account that we need
the preprocessor.  

what we could simplify is the need to explicitly set a compilation flag:
in a cabal file, one can use `if os(freebsd) ...` instead 
of `if flag(...)`, and be done with the extra with_freebsd flat.

dbalan:
@jaor I have updated the cabal file to use `if os(freebsd)` instead of an explicit flag. Much cleaner this way :)

jaor:
looks good, merged!

PR #423, opened 2020-01-10
Use a single Manager across the whole application

slotThe:
As per http-conduit's
[documentation](https://hackage.haskell.org/package/http-conduit-2.3.7.3/docs/Network-HTTP-Conduit.html):
"Creating a new Manager is a relatively expensive operation, you are advised to
share a single Manager between requests instead."

So far, every monitor that needed to connect to the internet (`Weather` and
`UVMeter`) created a new manager every time a connection was made.  This commit
introduces being able to set a manager via a record field in the `Monitor` type, so
that it can be retrieved by applications at will.

A caveat of this is that we now have a dependency on `http-conduit`, even if
neither of the plugins are loaded.  I would imagine this could be solved via an
`ifdef`, but I did not want to use the preprocessor any more than absolutely
necessary.  It's your call of course.

### DIFF
--- a/src/Xmobar/Plugins/Monitors/Common/Types.hs
+++ b/src/Xmobar/Plugins/Monitors/Common/Types.hs
@@ -25,8 +25,9 @@ module Xmobar.Plugins.Monitors.Common.Types ( Monitor
                                             , io
                                             ) where
 
-import Data.IORef
-import Control.Monad.Reader
+import Control.Monad.Reader (ReaderT, ask, liftIO)
+import Data.IORef (IORef, modifyIORef, newIORef, readIORef)
+
 
 type Monitor a = ReaderT MConfig IO a
 
--- a/changelog.md
+++ b/changelog.md
@@ -14,6 +14,9 @@ _New features_
     `offString` (thanks to *slotThe*).
   - `Battery` and `BatteryN` now support FreeBSD (thanks to Dhananjay
     Balan).
+  - New option `--useManager` for `Weather` and `UVMeter` to decide whether to
+    use one single manager per monitor for managing network connections or
+    create a new one every time a connection is made.
 
 ## Version 0.32 (December, 2019)
 
--- a/readme.md
+++ b/readme.md
@@ -740,6 +740,11 @@ something like:
     variable is not reported.
     - short option: `-w`
     - Default: ""
+  - `--useManager` _bool_ : Whether to use one single manager per monitor for
+    managing network connections or create a new one every time a connection is
+    made.
+    - Short option: `-m`
+    - Default: True
 - Variables that can be used with the `-t`/`--template` argument:
 	    `station`, `stationState`, `year`, `month`, `day`, `hour`,
 	    `windCardinal`, `windAzimuth`, `windMph`, `windKnots`, `windMs`, `windKmh`
@@ -1414,9 +1419,14 @@ following differences:
 
 ### `UVMeter`
 
-- Aliases to "uv " + station id. For example: `%uv brisbane%` or `%uv
-  alice springs%`
-- Args: default monitor arguments.
+- Aliases to "uv " + station id. For example: `%uv Brisbane%` or `%uv
+  Alice Springs%`
+- Args: default monitor arguments, plus:
+  - `--useManager` _bool_ : Whether to use one single manager per monitor for
+    managing network connections or create a new one every time a connection is
+    made.
+    - Short option: `-m`
+    - Default: True
 
 - *Reminder:* Keep the refresh rate high, to avoid making unnecessary
   requests every time the plug-in is run.
@@ -1424,7 +1434,7 @@ following differences:
   http://www.arpansa.gov.au/uvindex/realtime/xml/uvvalues.xml
 - Example:
 
-        Run UVMeter "brisbane" ["-H", "3", "-L", "3", "--low", "green", "--high", "red"] 900
+        Run UVMeter "Brisbane" ["-H", "3", "-L", "3", "--low", "green", "--high", "red"] 900
 
 ## Executing External Commands
 
--- a/src/Xmobar/Plugins/Monitors.hs
+++ b/src/Xmobar/Plugins/Monitors.hs
@@ -164,8 +164,8 @@ instance Exec Monitors where
     start (TopProc a r) = startTop a r
     start (TopMem a r) = runM a topMemConfig runTopMem r
 #ifdef WEATHER
-    start (Weather s a r) = runMD (s : a) weatherConfig runWeather r weatherReady
-    start (WeatherX s c a r) = runMD (s : a) weatherConfig (runWeather' c) r weatherReady
+    start (Weather  s   a r) = startWeather    s a r
+    start (WeatherX s c a r) = startWeather' c s a r
 #endif
     start (Thermal z a r) = runM (a ++ [z]) thermalConfig runThermal r
     start (ThermalZone z a r) =
--- a/src/Xmobar/Plugins/Monitors.hs
+++ b/src/Xmobar/Plugins/Monitors.hs
@@ -19,7 +19,7 @@ module Xmobar.Plugins.Monitors where
 
 import Xmobar.Run.Exec
 
-import Xmobar.Plugins.Monitors.Common (runM, runMD)
+import Xmobar.Plugins.Monitors.Common (runM)
 #ifdef WEATHER
 import Xmobar.Plugins.Monitors.Weather
 #endif
@@ -47,7 +47,7 @@ import Xmobar.Plugins.Monitors.Wireless
 #endif
 #ifdef LIBMPD
 import Xmobar.Plugins.Monitors.MPD
-import Xmobar.Plugins.Monitors.Common (runMBD)
+import Xmobar.Plugins.Monitors.Common (runMBD, runMD)
 #endif
 #ifdef ALSA
 import Xmobar.Plugins.Monitors.Volume
@@ -184,7 +184,7 @@ instance Exec Monitors where
     start (Uptime a r) = runM a uptimeConfig runUptime r
     start (CatInt _ s a r) = runM a catIntConfig (runCatInt s) r
 #ifdef UVMETER
-    start (UVMeter s a r) = runM (a ++ [s]) uvConfig runUVMeter r
+    start (UVMeter s a r) = startUVMeter s a r
 #endif
 #ifdef IWLIB
     start (Wireless i a r) = runM a wirelessConfig (runWireless i) r
--- a/src/Xmobar/Plugins/Monitors/Common/Run.hs
+++ b/src/Xmobar/Plugins/Monitors/Common/Run.hs
@@ -19,6 +19,7 @@ module Xmobar.Plugins.Monitors.Common.Run ( runM
                                           , runMD
                                           , runMB
                                           , runMBD
+                                          , getArgvs
                                           ) where
 
 import Control.Exception (SomeException,handle)
@@ -54,6 +55,13 @@ options =
     , Option "T" ["maxtwidth"] (ReqArg MaxTotalWidth "Maximum total width") "Maximum total width"
     , Option "E" ["maxtwidthellipsis"] (ReqArg MaxTotalWidthEllipsis "Maximum total width ellipsis") "Ellipsis to be added to the total text when it has reached its max width."
     ]
+
+-- | Get all argument values out of a list of arguments.
+getArgvs :: [String] -> [String]
+getArgvs args =
+    case getOpt Permute options args of
+        (_, n, []  ) -> n
+        (_, _, errs) -> errs
 
 doArgs :: [String]
        -> ([String] -> Monitor String)
--- a/src/Xmobar/Plugins/Monitors/CoreTemp.hs
+++ b/src/Xmobar/Plugins/Monitors/CoreTemp.hs
@@ -25,8 +25,7 @@ import Data.Char (isDigit)
 coreTempConfig :: IO MConfig
 coreTempConfig = mkMConfig
        "Temp: <core0>C" -- template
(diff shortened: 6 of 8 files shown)


jaor:
we can make the inclusion of http-conduits on compiling either the
weather plugin or uvmeter: both are controllable with flags in
xmobar.cabal (in the case of weather, precisely to avoid bringing in a
huge dep)... the flags are `with_weather` and `with_uvmeter`, we don't
need any extra `#ifdefs` :)

jaor:
oh... we possibly need an `#ifdef` or two, but since this dependency
is exclusively needed by monitors that are opt-out (uvmeter is even
out of `all_extensions`), i'd rather have it also opt-out.

slotThe:
Okay, I've added (hopefully all of) the necessary `#ifdef`s

jaor:
excellent, thanks for your patience!

i've rebased your two commits and pushed to master.  closing this pr.

jaor:
@slotThe, something seems wrong with this.  Using this shared monitor (as opposed to 2 commits before), my xmobar is consuming ~20% of a single CPU, while before this change it was never above 2-3%.  Do you see anything of the sort? (i'm using the weather monitor).

slotThe:
That is very weird indeed, which refresh time do you use?

Setting the time to something short-ish (like 10 seconds), without the
change I get a spike of around 11-15% CPU usage when the weather monitor gets
refreshed.  With the change I get 6-9% instead.  This seems to be the exact
opposite behaviour you are experiencing...

I can measure a performance decrease if I set the refresh time to something
very short, like 1ms.  Though the problem seems to go away if I remove the
`IORef` from the `manager` field (it's not needed anyway as we are not writing
to it).  Limiting open connections to 1 also seems to help somewhat (for
whatever reason).

I admit that my methodology is not exactly sound (basically staring at `htop` a
lot in order to see spikes and recording some averages with `atop`) and I will
try to go over this a little bit more carefully when I have the time.

@jaor in the meantime can you try if this new commit changes anything for you?
It just removes the `IORef` from the `manager` field and sets the limit for
open connections to one.

jaor:
On Mon, Jan 27 2020, slotThe wrote:

> That is very weird indeed, which refresh time do you use?

half an hour (18000 in the config).

> Setting the time to something short-ish (like 10 seconds), without the
> change I get a spike of around 11-15% CPU usage when the weather monitor gets
> refreshed. With the change I get 6-9% instead. This seems to be the exact
> opposite behaviour you are experiencing...

indeed.  what i see is a pretty consistent CPU occupation of around 20%,
no spikes: it doesn't seem caused by network access.

> I can measure a performance decrease if I set the refresh time to something
> very short, like 1ms. Though the problem seems to go away if I remove the
> IORef from the manager field (it's not needed anyway as we are not writing
> to it). Limiting open connections to 1 also seems to help somewhat (for
> whatever reason).
>
> I admit that my methodology is not exactly sound (basically staring at htop a
> lot in order to see spikes and recording some averages with atop) and I will
> try to go over this a little bit more carefully when I have the time.

i use xmobar itself (the Top monitor), not specially sound either :)

> @jaor in the meantime can you try if this new commit changes anything
> for you?  It just removes the IORef from the manager field and sets
> the limit for open connections to one.

no, same behaviour.  in case that helps, my template looks like

    "<skyConditionS> <tempC>° <rh>% <windKmh> (<hour>)"

and i'm setting ["-w", ""] explicitly in the additional options
(although i don't see how that would have an effect).  i'm a bit busy
right now, but i guess i can try with simpler things to see if i can
reproduce the problem with less convoluted configurations (i use now a
pure haskell one https://gitlab.com/jaor/xmobar-config/, this is with
xmobar-exwm).

slotThe:
> no, same behaviour. in case that helps, my template looks like "<skyConditionS> <tempC>° <rh>% <windKmh> (<hour>)" and i'm setting ["-w", ""] explicitly in the additional options (although i don't see how that would have an effect). i'm a bit busy right now, but i guess i can try with simpler things to see if i can reproduce the problem with less convoluted configurations (i use now a pure haskell one https://gitlab.com/jaor/xmobar-config/, this is with xmobar-exwm).

Okay, so I can actually reproduce this with your config (I compiled and used
the `xmobar-exwm` executable).  Exact same behaviour, constant 20%-ish cpu
usage.  However, copying the relevant weather part over to my config (for
reference, available [here](https://gitlab.com/slotThe/dotfiles/tree/master/xmobar/.config/xmobarrc)) makes the problem go away.  This seems to be some
weird config-assisted madness, and I would very much be interested in a minimal
example of this behaviour, should you have the time to hunt for it.

jaor:
On Tue, Jan 28 2020, slotThe wrote:

> Okay, so I can actually reproduce this with your config (I compiled and used
> the xmobar-exwm executable). Exact same behaviour, constant 20%-ish cpu
> usage. However, copying the relevant weather part over to my config (for
> reference, available here) makes the problem go away. This seems to be some
> weird config-assisted madness, and I would very much be interested in a minimal
> example of this behaviour, should you have the time to hunt for it.

i've played a little and it's not the weather specifically.  removing it
still makes xmobar-exwm consume more CPU cycles.  the additional
activity seems evenly distributed across monitor: if one enables one by
one, CPU usages increases steadily until that 20%, and i cannot see one
of them causing a spike... so, if one runs something minimal, the bad
behaviour is not as evident, but it seems to still be there...

slotThe:
> looking again at the code, i see we're creating a newManager every
> time mkMConfig is called. now, _all monitors_ create an instance of
> mkMConfig, so unless newManager is returning always the same instance
> (which, given its name, would be surprising), i think we're creating
> lots of (unused) managers: am i missing something here?

Whoops, you are right.  That's on me, sorry.  I've pushed a quick fix of this issue (basically adding another argument to `mkMConfig`), though I'm not quite sure I like this solution.

> (also, not sure if this is the cause the CPU problem...)

I have just tried it with your config again, but sadly that didn't seem to be the issue.

jaor:
On Wed, Jan 29 2020, slotThe wrote:

[...]

> I have just tried it with your config again, but sadly that didn't
> seem to be the issue.

i'm guessing then that somehow creating managers produces this effect (i
supposed that, with this new version, not using weather in the config
makes the problem disappear?)

btw, i was reading the doc of http-conduit, and about managers it
mentions that they're useful in keep-alive situations... for the weather
monitor, which typically is updated every half an hour, there doesn't
seem to be much need for a manager: sorry if you already mentioned this
and i forgot, but do you have other use cases in mind?

slotThe:
> i'm guessing then that somehow creating managers produces this effect (i supposed that, with this new version, not using weather in the config makes the problem disappear?) 

Indeed, though now that I test it again, it seems like the problem is gone even if I enable the weather monitor again.  It might be that this was the cause of the issue after all and I just weren't on the right branch when testing this earlier.  Can you confirm?  Sorry for this mess, I'm running on very little sleep right now.

>btw, i was reading the doc of http-conduit, and about managers it mentions that they're useful in keep-alive situations... for the weather monitor, which typically is updated every half an hour, there doesn't seem to be much need for a manager: sorry if you already mentioned this and i forgot, but do you have other use cases in mind?

The purpose of the manager in general is keep-alive yes.  However, I thought it a waste of resources to create a new manager every time one wants to make a request when it's perfectly fine to create it once instead.  This is especially true because we are not contacting very many hosts, but the same one over and over again (not to mention people who really want to know whether the weather changed in the last 10 seconds :>).  It definitely worked for me (and with this latest commit now hopefully for everyone else as well), as I can see much less cpu usage when updating these monitors now.

jaor:
On Wed, Jan 29 2020, slotThe wrote:

> The purpose of the manager in general is keep-alive yes. However, I
> thought it a waste of resources to create a new manager every time one
> wants to make a request when it's perfectly fine to create it once
> instead. This is especially true because we are not contacting very
> many hosts, but the same one over and over again (not to mention
> people who really want to know whether the weather changed in the last
> 10 seconds :>). It definitely worked for me (and with this latest
> commit now hopefully for everyone else as well), as I can see much
> less cpu usage when updating these monitors now.

i see. i just checked, and now the increase in cpu consumption is hardly
noticeable.  it's still there though, and it bothers me a bit that we
have to pay this price... after all, using the manager is supposed to be
a resource optimization, and turns out that is making xmobar itself
consuming more power.  how about we activate the persistent manager only
when refresh time is less than, say 10 secs?

anyway, regarding the patch as it stands, if one compiles without
weather, we'll get a warning because the mngr parameter in mkMConfig is
unused.  could we include the signature of the function in the #if
condition to avoid that?

(sorry for being such a nitpicker! get some sleep first please :))

src/Xmobar/Plugins/Monitors/Common/Types.hs
```diff
@@ -102,7 +120,20 @@ mkMConfig tmpl exprts =
...
+#if defined(WEATHER) || defined(UVMETER)
+       man <- if mngr
+           then newIORef =<< makeNewManager
+           else newIORef Nothing
+       return $ MC nc l lc h hc t e p d mn mx mel pc pr bb bf bw up na mt mtel man
+
+-- | Create a new 'Manager' for managing network connections.
+makeNewManager :: IO (Maybe Manager)
+makeNewManager =
+    Just <$> newManager (tlsManagerSettings { managerConnCount = 1 })
```
jaor: this way of handling the manager, with a boolean that is almost always false and that actually doesn't even make sense if compiled without weather support is a bit ugly.  and moreover we're putting the manager in a ioRef, so it's not even for the sake of keeping the code pure.

how about the relevant plugins just declare a newIORef at their top level, and fill it in with makeNewManager as needed, leaving the configuration for all monitors alone?  it's also not very natural to think of this manager as a configuration parameter.

that would reduce the overall complexity of the code, and also make this PR simpler.

slotThe:
>i see. i just checked, and now the increase in cpu consumption is hardly noticeable.  it's still there though, and it bothers me a bit that we have to pay this price... after all, using the manager is supposed to be a resource optimization, and turns out that is making xmobar itself consuming more power.

This worries me a bit, I will try to investigate this at some point.


Okay, after a good nights sleep I can now see how horrible that Bool solution
really was.  I can think of two ways to proceed:

  1. Keep the field in `MConfig`, but set it to `Nothing` by default and fill
     it in inside `weatherReady` and UVMeter's `getData` respectively.

  2. Delete the field from `MConfig` and make it a local `IORef` for the
     monitors that need it.  I really don't want to create a new `run` like
     function for this (I think we have enough of those as it is), so this would
     involve creating two managers for monitors that have a `ready` function.

While 2. would involve creating two managers in the case of `Weather`, it would
also enable us to get rid of all the `#if`'s.  This makes me kinda wanna go
down that route. What do you think?

> how about we activate the persistent manager only when refresh time is less than, say 10 secs?

I would personally be in favour of making this a configurable option (for
people like me who update the weather monitor less frequently, but where xmobar
doesn't seem to consume more power with this).  Is that okay for you as well?

jaor:
On Fri, Jan 31 2020, slotThe wrote:

>     i see. i just checked, and now the increase in cpu consumption is
>     hardly noticeable. it's still there though, and it bothers me a
>     bit that we have to pay this price... after all, using the manager
>     is supposed to be a resource optimization, and turns out that is
>     making xmobar itself consuming more power.
>
> This worries me a bit, I will try to investigate this at some point.

thanks.  while i think it's not critical (with just one monitor using it
the effect is small), it'd be nice to understand what's going on.

> Okay, after a good nights sleep I can now see how horrible that Bool solution
> really was. I can think of two ways to proceed:
>
>  1. Keep the field in MConfig, but set it to Nothing by default and fill
>     it in inside weatherReady and UVMeter's getData respectively.
>
>  2. Delete the field from MConfig and make it a local IORef for the
>     monitors that need it. I really don't want to create a new run like
>     function for this (I think we have enough of those as it is), so this would
>     involve creating two managers for monitors that have a ready function.
>
> While 2. would involve creating two managers in the case of Weather,
> it would also enable us to get rid of all the #if's. This makes me
> kinda wanna go down that route. What do you think?

2 sounds good to me too.  i like a lot that we get rid of the #ifs, and
i don't see a big problem with the two managers.  as i mentioned
elsewhere, having the manager in config feels awkward to me (what i
would expect there, if anything, would be a useManager boolean flag, as
you anticipate below).

>     how about we activate the persistent manager only when refresh
>     time is less than, say 10 secs?
>
> I would personally be in favour of making this a configurable option (for
> people like me who update the weather monitor less frequently, but where xmobar
> doesn't seem to consume more power with this). Is that okay for you as well?

that would be perfect, yes.  what do you think of something like the
useManager that i mention above?  i'm fine with having it set to True by
default, and people experiencing any kind of trouble, or optimizing for
battery time, would have a escape hatch.

thanks!
jao
--
Simplicity does not precede complexity, but follows it.
  - Alan Perlis, Epigrams on Programming

slotThe:
> that would be perfect, yes. what do you think of something like the useManager that i mention above?

Sounds good to me, these last few commits should achieve exactly that!  I did
however end up implementing this as a per-monitor setting.  It would have been
nice for this to be a new field for the actual `Config` type, but from looking
at it the config is processed before the monitors come into play at all, so
this would have probably required some rewriting of the core of xmobar (I
deemed this very much unnecessary for just two monitors, both of which aren't
even enabled by default).

Some caveats of this solution are that we are defining the same two functions
(`tryMakeManager` and `mkManager`) in `Weather.hs` and `UVMeter.hs` (though
this again saves `#if`'s) and that we are parsing the options for the two
monitors one additional time, though since they seem to get parsed every time
they update this seems negligible.

jaor:
slotThe, i cannot merge cleanly this PR: would you please rebase it on
top of master? thanks!

slotThe:
> slotThe, i cannot merge cleanly this PR: would you please rebase it on top of master? thanks!

Should be all done!

jaor:
merged, thanks!

PR #427, opened 2020-02-04
fix --high to --highs in documentation

eyenx:
`--high` does not work. `--highs` is the correct parameter in BatteryP


jaor:
thanks!

PR #428, opened 2020-02-11
Battery: fix time calculations on some Linux systems

paulfertser:
The calculations are based on the assumption that current_now (or
power_now) are always positive. However, power_supply documentation in
the kernel sources say nothing about it, and so some drivers provide a
signed value (e.g. bq27xxx_battery_current).

Discovered and fixed on ac100/paz00 netbook.


jaor:
merged, thanks!

PR #429, opened 2020-02-17
Wireless: support NL80211 userspace <-> kernelspace API

paulfertser:
NL80211 was introduced in Linux 2.6.24 in 2007 as a new extensible
universal API, replacing "wireless extensions" ioctls. It works on top
of netlink, and allows direct communication to cfg80211 kernel
subsystem. Since then it became a hard requirement for all upstream
wireless drivers to hook into cfg80211 (SoftMAC drivers do it via the
common mac80211 layer). There's still additional compatibility code that
allows limited Wext functionality for cfg80211 drivers but it's buggy
and can be disabled altogether when CONFIG_CFG80211_WEXT is not set.

This patch makes use of "netlink" Haskell library which doesn't have any
additional runtime dependencies (so neither iwlib nor libnl are
required). The operation is the same as used by "iw dev <devname> link"
command.

I decided against trying to map the reported signal to any kind of
"quality"/percentages/"bars" after talking to some Linux wireless
developers and WiFi users. The reason is that it's not possible to
predict throughput even for a specific card and antenna combination
knowing just the signal level (without SNR, channel busy time etc).
Signal level alone can be somewhat useful when you need to know if
you're moving closer or farther from an AP with unknown location or
while positioning a directional antenna, and for these purposes raw dBm
numbers are most useful.

I didn't yet describe the functionality in the readme, can do that along with the next patch version after getting feedback.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -20,6 +20,8 @@ _New features_
   - New more efficient time coalescing strategy for monitor updates,
     available with the threaded runtime: use the `with_threaded` flag
     to enable it (see #410; thanks to Tomáš Janoušek).
+  - `Wireless` supports current nl80211 API on Linux now, old Wext ioctls still
+    available as compile-time option (thanks to Paul Fertser).
 
 ## Version 0.32 (December, 2019)
 
--- a/readme.md
+++ b/readme.md
@@ -117,11 +117,15 @@ Otherwise, you'll need to install them yourself.
   option is needed for the MBox and Mail plugins to work. Requires the
   [hinotify] package.
 
-- `with_iwlib` Support for wireless cards. Enables the Wireless
-   plugin. No Haskell library is required, but you will need the
-   [iwlib] C library and headers in your system (e.g., install
-   `libiw-dev` in Debian-based systems or `wireless_tools` on Arch
-   Linux).
+- `with_nl80211` Support for wireless cards on Linux via nl80211 (all
+   upstream drivers). Enables the Wireless plugin. Requires [netlink]
+   and [cereal] packages.
+
+- `with_iwlib` Support for wireless cards via Wext ioctls
+   (deprecated). Enables the Wireless plugin. No Haskell library is
+   required, but you will need the [iwlib] C library and headers in your
+   system (e.g., install `libiw-dev` in Debian-based systems or
+   `wireless_tools` on Arch Linux). Conflicts with `with_nl80211`.
 
 - `with_alsa` Support for ALSA sound cards. Enables the Volume
    plugin. Requires the [alsa-mixer] package.  To install the latter,
@@ -820,18 +824,21 @@ specification, such as `("clear", "<icon=weather-clear.xbm/>")`.
 
 ### `Wireless Interface Args RefreshRate`
 
-- If set to "", the interface is looked up in /proc/net/wireless.
+- If set to "", first suitable wireless interface is used.
 - Aliases to the interface name with the suffix "wi": thus, `Wireless
   "wlan0" []` can be used as `%wlan0wi%`, and `Wireless "" []` as `%wi%`.
 - Args: default monitor arguments, plus:
   - `--quality-icon-pattern`: dynamic string for connection quality in `qualityipat`.
 - Variables that can be used with the `-t`/`--template` argument:
-            `essid`, `quality`, `qualitybar`, `qualityvbar`, `qualityipat`
-- Thresholds refer to link quality in a `[0, 100]` scale
-- Default template: `<essid> <quality>`
-- Requires the C library [iwlib] (part of the wireless tools suite)
-  installed in your system. In addition, to activate this plugin you
-  must pass `--flags="with_iwlib"` during compilation
+            `ssid`, `signal`, `quality`, `qualitybar`, `qualityvbar`, `qualityipat`
+- Thresholds refer to link quality on a `[0, 100]` scale. Note that
+  `quality` is calculated from `signal` (in dBm) by a possibly lossy
+  conversion. It is also not taking into account many factors such as
+  noise level, air busy time, transcievers' capabilities and the
+  others which can have drastic impact on the link performance.
+- Default template: `<ssid> <quality>`
+- To activate this plugin you must pass `--flags="with_nl80211"` or
+  `--flags="with_iwlib"` during compilation
 
 ### `Memory Args RefreshRate`
 
--- a/src/Xmobar/Plugins/Monitors.hs
+++ b/src/Xmobar/Plugins/Monitors.hs
@@ -42,7 +42,7 @@ import Xmobar.Plugins.Monitors.CatInt
 #ifdef UVMETER
 import Xmobar.Plugins.Monitors.UVMeter
 #endif
-#ifdef IWLIB
+#if defined IWLIB || defined USE_NL80211
 import Xmobar.Plugins.Monitors.Wireless
 #endif
 #ifdef LIBMPD
@@ -85,7 +85,7 @@ data Monitors = Network      Interface   Args Rate
 #ifdef UVMETER
               | UVMeter      Station     Args Rate
 #endif
-#ifdef IWLIB
+#if defined IWLIB || defined USE_NL80211
               | Wireless Interface  Args Rate
 #endif
 #ifdef LIBMPD
@@ -142,7 +142,7 @@ instance Exec Monitors where
 #ifdef UVMETER
     alias (UVMeter s _ _) = "uv " ++ s
 #endif
-#ifdef IWLIB
+#if defined IWLIB || defined USE_NL80211
     alias (Wireless i _ _) = i ++ "wi"
 #endif
 #ifdef LIBMPD
@@ -186,7 +186,7 @@ instance Exec Monitors where
 #ifdef UVMETER
     start (UVMeter s a r) = startUVMeter s a r
 #endif
-#ifdef IWLIB
+#if defined IWLIB || defined USE_NL80211
     start (Wireless i a r) = runM a wirelessConfig (runWireless i) r
 #endif
 #ifdef LIBMPD
--- a/src/Xmobar/Plugins/Monitors/Wireless.hs
+++ b/src/Xmobar/Plugins/Monitors/Wireless.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE TypeApplications, CPP #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  Plugins.Monitors.Wireless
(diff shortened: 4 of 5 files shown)


jaor:
thanks for working on this, it's nice to get rid of non-haskell deps.
and thanks also for the thorough explanation.

the code looks good, but i'm worried about not providing a [0-100]
quality measure.  for many years, the wireless monitor has done just
that and users are going to be surprised by such a drastic change in
behaviour (more so because netlink becomes now the default)... even
though i understand your concerns, right now i'm seeing -37.6 in my
xmobar, and that number conveys almost no information to me, and i'm
afraid many users would feel the same.  even raw iwconfig is giving a
"link quality" as a quotient.  couldn't we just have both, `<quality>`
and `<level>`? (at any rate, calling `<quality>` the signal level
sounds wrong).

what do you think?

paulfertser:
Hello

Thank you for the feedback.

On Sat, Feb 22, 2020 at 02:49:52PM -0800, jao wrote:
> right now i'm seeing -37.6 in my xmobar, and that number conveys
> almost no information to me, and i'm afraid many users would feel
> the same.

Are you sure it's -37.6 and not -37.0? The value returned from
NL80211_CMD_GET_STATION is a signed 8-bit integer (in dBm) and so
there can be no fractional part.

My notion is that there's really almost no useful information in the
values you can get from the wireless drivers. Their signal level
measurements are not supposed to be accurate absolute numbers (and
some cards (e.g. ath10k[0]) do not even provide it at all, they give
SNR instead which the drivers somehow "convert" to signal level to fit
the mac80211 framework), and in any case you do not have information
about both signal level and SNR. So there's not much to expect.

One might consider the Tx/Rx bitrates to be indicators of the link
quality (because the driver + card should be trying to get the maximum
given everything they know about the medium conditions) but that's
true only for the fully loaded links as when there's no traffic most
equipment lowers the bitrate to get more reliability (you can observe
it all by monitoring "iw dev ... link" output in different conditions).

> even raw iwconfig is giving a "link quality" as a quotient.

iwconfig is using the deprecated wireless extensions ioctls. They're
handled by the CFG80211 compatibility layer on your system. It does a
trivial (and sometimes data losing) mapping[0]. -37.0 in your case
would mean it first gets clamped to -40, then 110 added and you end up
seeing 70. Does this match the behaviour you observe?

I can't see how this can be useful anyhow.

> couldn't we just have both, <quality> and <level>? (at any rate,
> calling <quality> the signal level sounds wrong).

I have a feeling we can't have quality because there's no way to learn
it. I asked few wifi users and they all said "quality" indicators in
their devices are ~useless (and I'm not surprised). Having signal
level indicator can be meaningful for positioning directional antennas
or for "radio fox hunting", and in both cases I think no conversion is
better than any conversion.

Did you have any positive experience with "quality" indicators?

[0] https://git.kernel.org/pub/scm/linux/kernel/git/wireless/wireless-testing.git/tree/drivers/net/wireless/ath/ath10k/wmi.c#n2527
[1] https://git.kernel.org/pub/scm/linux/kernel/git/wireless/wireless-testing.git/tree/net/wireless/wext-compat.c#n1331
-- 
Be free, use free (http://www.gnu.org/philosophy/free-sw.html) software!
mailto:fercerpav@gmail.com

jaor:
hi paul,

On Sat, Feb 22 2020, Paul Fertser wrote:

> Hello
>
> Thank you for the feedback.
>
> On Sat, Feb 22, 2020 at 02:49:52PM -0800, jao wrote:
>> right now i'm seeing -37.6 in my xmobar, and that number conveys
>> almost no information to me, and i'm afraid many users would feel
>> the same.
>
> Are you sure it's -37.6 and not -37.0? The value returned from

yes, sorry, fat fingers :)

> NL80211_CMD_GET_STATION is a signed 8-bit integer (in dBm) and so
> there can be no fractional part.

(we should then eliminate the .0 altogether, maybe i'm asking for it in
my config.)

[...]

> iwconfig is using the deprecated wireless extensions ioctls. They're
> handled by the CFG80211 compatibility layer on your system. It does a
> trivial (and sometimes data losing) mapping[0]. -37.0 in your case
> would mean it first gets clamped to -40, then 110 added and you end up
> seeing 70. Does this match the behaviour you observe?

yes

> I have a feeling we can't have quality because there's no way to learn
> it. I asked few wifi users and they all said "quality" indicators in
> their devices are ~useless (and I'm not surprised). Having signal
> level indicator can be meaningful for positioning directional antennas
> or for "radio fox hunting", and in both cases I think no conversion is
> better than any conversion.
>
> Did you have any positive experience with "quality" indicators?

from what i'm learning thanks to your explanations, maybe not, but i
hadn't noticed: when i move away from my router the "quality" drops, and
i experience worse connectivity when it's low: that might well be a
subjective impression, but one that, i daresay, is shared by many of our
users.

while i might agree with you that the "quality" the iwlib monitor is
reporting is nearly useless, replacing it by a level is confusing.  i
think that if we're reporting level, we should call it level, not
quality.  and if the latter is not available for this new monitor
implementation, report "N/A" for it, or even don't provide it at all as
a template variable.

but if we follow that strategy, i wouldn't exclude by default the iwlib
flag from all_extensions, at least for now.  i'd make netlink an opt-in
that, when given explicitly (say with `with_netlink`), removes iwlib
from the used flags (even if all_extensions was also specified).  what
do you think? an alternative is to provide a separate wireless monitor
that simply doesn't provide "quality", with a different name, and
include both in all_extensions...

paulfertser:
On Sun, Feb 23, 2020 at 11:34:35AM -0800, jao wrote:
> > NL80211_CMD_GET_STATION is a signed 8-bit integer (in dBm) and so
> > there can be no fractional part.
> 
> (we should then eliminate the .0 altogether, maybe i'm asking for it in
> my config.)

Done.

> > Did you have any positive experience with "quality" indicators?
> 
> from what i'm learning thanks to your explanations, maybe not, but i
> hadn't noticed: when i move away from my router the "quality" drops, and
> i experience worse connectivity when it's low: that might well be a
> subjective impression, but one that, i daresay, is shared by many of our
> users.

So you're in a known location with relatively stable noise level and
known equipment, in this case the signal level would of course
correlate with the maximum speed and packet loss etc. But knowing
where you are and what the conditions and your previous experience are
you already know what results throughput-wise to expect, so in this
case I can't see how indicating signal level helps.

> i think that if we're reporting level, we should call it level, not
> quality.

Done.

> but if we follow that strategy, i wouldn't exclude by default the iwlib
> flag from all_extensions, at least for now. i'd make netlink an opt-in
> that, when given explicitly (say with `with_netlink`), removes iwlib
> from the used flags (even if all_extensions was also specified). what
> do you think? an alternative is to provide a separate wireless monitor
> that simply doesn't provide "quality", with a different name, and
> include both in all_extensions...

Oh, by that you've threatened me enough, see the updated code :)

-- 
Be free, use free (http://www.gnu.org/philosophy/free-sw.html) software!
mailto:fercerpav@gmail.com

jaor:
On Mon, Feb 24 2020, Paul Fertser wrote:

[...]

> So you're in a known location with relatively stable noise level and
> known equipment, in this case the signal level would of course
> correlate with the maximum speed and packet loss etc. But knowing
> where you are and what the conditions and your previous experience are
> you already know what results throughput-wise to expect, so in this
> case I can't see how indicating signal level helps.

sometimes my router will have issues and provide a lower level.
sometimes i'm visiting a friend with a big house, and discovering which
rooms there have no good coverage and by what (relative) amount...

>> but if we follow that strategy, i wouldn't exclude by default the iwlib
>> flag from all_extensions, at least for now. i'd make netlink an opt-in
>> that, when given explicitly (say with `with_netlink`), removes iwlib
>> from the used flags (even if all_extensions was also specified). what
>> do you think? an alternative is to provide a separate wireless monitor
>> that simply doesn't provide "quality", with a different name, and
>> include both in all_extensions...
>
> Oh, by that you've threatened me enough, see the updated code :)

sorry, i didn't mean to sound threatening! :) thanks for bearing with
me, the patch looks good... i'm testing and, if you don't do it before,
i'll rebase or cherry pick it on top of current master.

paulfertser:
On Mon, Feb 24, 2020 at 03:47:36AM -0800, jao wrote:
> sometimes my router will have issues and provide a lower level.
> sometimes i'm visiting a friend with a big house, and discovering which
> rooms there have no good coverage and by what (relative) amount...

Yeah, and signal level in dBm is just as useful as the "quality". In
any case, I have a feeling we're on the same page now regarding this,
our understanding is in sync.

> > Oh, by that you've threatened me enough, see the updated code :)
> 
> sorry, i didn't mean to sound threatening! :)

Well, your idea of making nl80211 non default or even a separate
monitor convinced (or, in other words, made) me add the conversion to
the patch, so here it is.

> i'm testing and, if you don't do it before, i'll rebase or cherry
> pick it on top of current master.

Please just tell me how your testing goes (and whether you agree to
deprecating "essid" in favour of "ssid" etc), and I'll rebase on top
of master and add relevant documentation changes while at it.

-- 
Be free, use free (http://www.gnu.org/philosophy/free-sw.html) software!
mailto:fercerpav@gmail.com

jaor:
On Mon, Feb 24 2020, Paul Fertser wrote:

[...]

>> i'm testing and, if you don't do it before, i'll rebase or cherry
>> pick it on top of current master.
>
> Please just tell me how your testing goes (and whether you agree to
> deprecating "essid" in favour of "ssid" etc), and I'll rebase on top
> of master and add relevant documentation changes while at it.

testing is going well so far, and i'm fine with the deprecation and the
rest of the patch, so please go on.  if you're taking the time to add
documentation, and feel like it, a brief note somewhere warning about
how quality is not a reliable metric and we would recommend using level
instead would be welcome (but please don't feel obliged if it's a chore,
we can always point to the discussion here).

thanks!

paulfertser:
Thank you very much for the feedback.

On Mon, Feb 24, 2020 at 11:31:15AM -0800, jao wrote:
> > deprecating "essid" in favour of "ssid" etc), and I'll rebase on top
> > of master and add relevant documentation changes while at it.
> 
> testing is going well so far, and i'm fine with the deprecation and the
> rest of the patch, so please go on. if you're taking the time to add
> documentation, and feel like it, a brief note somewhere warning about
> how quality is not a reliable metric and we would recommend using level
> instead would be welcome (but please don't feel obliged if it's a chore,
> we can always point to the discussion here).

I plan to do that tomorrow. BTW, if you feel like changing my code to
make it more readable/error-resistant/elegant/concise/whatever, please
feel free to; my Haskell (and FP in general) experience is very
limited and so I expect my solutions to be suboptimal. I was trying to
make minimal changes to the existing code so that IWLIB variant would
still be working exactly as it was before (but I haven't tested it).

-- 
Be free, use free (http://www.gnu.org/philosophy/free-sw.html) software!
mailto:fercerpav@gmail.com

jaor:
merged, thanks!

Issue #434, opened 2020-04-02
Net monitor showing NA for rx/tx stats


zevweiss:
After installing a recent xmobar on a new system, I found that the Net monitor only showed `NA` where I expected traffic stats.

After digging into it a bit I suspect this is probably really a problem with the bytestring package's `readFile` function not handling magic procfs files appropriately when `stat(2)` returns an `st_size` of zero, but I'm not really expert enough to say for sure.

Though I don't think it's really a "proper" fix, I was able to hack around it and get things working again with the following patch, for what it's worth:

```diff
diff --git a/src/Xmobar/Plugins/Monitors/Net.hs b/src/Xmobar/Plugins/Monitors/Net.hs
index 97deacdcb0f3..d3603ea25558 100644
--- a/src/Xmobar/Plugins/Monitors/Net.hs
+++ b/src/Xmobar/Plugins/Monitors/Net.hs
@@ -30,6 +30,7 @@ import Control.Monad (forM, filterM)
 import System.Directory (getDirectoryContents, doesFileExist)
 import System.FilePath ((</>))
 import System.Console.GetOpt
+import System.IO (withFile, IOMode(..))
 import System.IO.Error (catchIOError)
 import System.IO.Unsafe (unsafeInterleaveIO)
 
@@ -137,12 +138,13 @@ netParser = mapM (readNetDev . splitDevLine) . readDevLines
 
 findNetDev :: String -> IO NetDevRawTotal
 findNetDev dev = do
-  nds <- B.readFile "/proc/net/dev" >>= netParser
+  nds <- bReadFile "/proc/net/dev" >>= netParser
   case filter isDev nds of
     x:_ -> return x
     _ -> return NA
   where isDev (N d _) = d == dev
         isDev NA = False
+        bReadFile p = withFile p ReadMode B.hGetContents
 
 formatNet :: Maybe IconPattern -> Float -> Monitor (String, String, String, String)
 formatNet mipat d = do

```

zevweiss:
Ah, and after being unable to reproduce the problem on another system with a newer GHC, it appears the problem is in fact in the bytestring code and has already been fixed upstream in commit [7223f8f6d9](https://github.com/haskell/bytestring/commit/7223f8f6d921e4c2fb118de95e65ab9f27307238).  After tweaking xmonad.cabal to specify bytestring >= 0.10.8.2 (which contains the fix), I now have a working version.

PR #435, opened 2020-04-02
Require bytestring >= 0.10.8.2.

zevweiss:
Version 0.10.8.1 contains a bug in the readFile function that misbehaves
on things like magic procfs files where stat(2) returns an st_size of
zero, which breaks the Net monitor and such; 0.10.8.2 contains the fix.

(See #434.)


jaor:
merged. thanks a lot for looking into this.

PR #436, opened 2020-04-09
Add a HandleReader Plugin

prikhi:
Hiya! It's me, the human that initiated the library refactoring in #341. 

I finally got around to integrating the code you ended up merging into my XMonad config:
https://github.com/prikhi/xmonad-config/commit/ee175e7ec176163bda830748428b0fc1e319d62f

It worked great, but I wanted to get rid of the complexity of creating a named pipe, setting permissions correctly, and then passing it to xmobar's `PipeReader`. So I made a new plugin that simply reads from a Haskell `Handle` instead. This let me remove a lot more cruft from my config:
https://github.com/prikhi/xmonad-config/commit/770ec8634f930f7bb1626e873fc3abc8af0901fe#diff-d17b1c671161a65360103ca06cd5a3a9

I think it'd be great to share this with everyone, so now there's this PR!

Eventually, I'd like to get the xmobar process & handle management into `xmonad-contrib` or a new package to make it even easier to configure xmobar from within an xmonad but I think we're all good on the xmobar side of that.

Let me know if you have any tweaks, questions, concerns, or comments.

Commit message follows:

---

This adds a new `HandleReader` plugin, which displays data from a
Haskell `Handle`. This is really only useful if you are running xmobar
from within another Haskell program, but lets you avoid the mechanics of
creating a named pipe with the proper file permissions.

Instead, you can use `System.Process.createPipe` to make a pair of read
& write Handles. If you pass the read handle to HandleReader, you can
use hPutStr on the write Handle to send data to xmobar from your
application code.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,10 @@
+## Version 0.34 (Unreleased)
+
+_New features_
+
+  - New plugin `HandleReader` for reading data from a Haskell `Handle`. This is
+    useful if you are running xmobar from within a Haskell program.
+
 ## Version 0.33 (February, 2020)
 
 _New features_
--- a/readme.md
+++ b/readme.md
@@ -1613,6 +1613,23 @@ will display "N/A" if for some reason the `date` invocation fails.
           logHook = dynamicLogString myPP >>= xmonadPropLog
         }
 
+### `HandleReader Handle Alias`
+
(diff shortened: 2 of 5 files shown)


jaor:
this is also useful when using xmobar as a lib to compile your own xmobar exec (a la xmonad). merged!

PR #437, opened 2020-04-09
Fix Tests When Compiling Without ALSA

prikhi:
Use CPP macros to make the ALSA Spec file a no-op when the `with_alsa`
flag is set to False. Without these macros, HSpec will import the
AlsaSpec module causing test compilation to fail.


jaor:
thanks!

Issue #438, opened 2020-04-20
High CPU throttling of xmobar


psibi:
Versions:
* xmobar version: 0.33
* Linux kernel: 5.3.0

The high CPU throttling starts after  some minutes of starting the system (I can hear my CPU fans after some time). I can confirm by using `htop` that it is actually `xmobar` which is having high CPU usage. I disable various flags like weather, utf8 but even then the high CPU usage occurred. Let me know if more information is needed.

jaor:
well, your config, and os, perhaps?

for xmobar, a minimal one that reproduces the behaviour, removing as
many monitors as possible, would help.  also your window manager, its
version...

psibi:
This is my xmobar config:
```
Config { font = "xft:Ubuntu Mono:size=12:bold:antialias=true"
       , bgColor = "black"
       , fgColor = "grey"
       , position = TopW L 100
       , lowerOnStart = False
       , hideOnStart = False
       , overrideRedirect = True
       , allDesktops = True
       , commands = [
                      Run Cpu ["-L","3","-H","50","--normal","green","--high","red"] 10
                    , Run Memory ["-t","Mem: <usedratio>%"] 10
                    , Run Swap [] 10
                    , Run Battery ["-t","<left>","-L","30","-H","70","--low", "red","--normal","yellow","--high","green"] 10
                    , Run Date "%a %b %_d %l:%M" "date" 10
                    , Run StdinReader
                    ]
       , sepChar = "%"
       , alignSep = "}{"
       , template = "%StdinReader% }{ | %cpu% | %memory% * %swap% * %battery% | <fc=#ee9a00>%date%</fc>"
       }
```
OS is ubuntu although I'm managing my xmobar dependency via nix. Using xmonad `0.15` (again the `xmonad` is being managed via nix. There is only a single monitor. I tried to get an strace log while the CPU was in full throttle mode:
https://gist.github.com/psibi/f522b5d0b4647fbdf5e724b13df58294

Let me know if you need anything more.

psibi:
Also, I'm planning to spend more time debugging this the coming weekend. So, let me know if you have any suspicious code paths. I will try to start from there. Thanks!

jaor:
> This is my xmobar config:
>
> Config { font = "xft:Ubuntu Mono:size=12:bold:antialias=true"
>        , bgColor = "black"
>        , fgColor = "grey"
>        , position = TopW L 100
>        , lowerOnStart = False
>        , hideOnStart = False
[...]

PR #439, opened 2020-04-25
Avoid busy looping by not catching all exceptions

psibi:
This specifically avoids situation described in this issue https://github.com/jaor/xmobar/issues/438 where the handle was throwing the IOException continously in a loop:

```
<stdin>: hGetLine: invalid argument (invalid byte sequence)
```

It happened because my system's environment wasn't right, but the proper behaviour here would be to let it to throw the exception rather than leading to a busy loop.

I did some git blame to find out that this commit introduced the behaviour: https://github.com/jaor/xmobar/commit/fc24dc1874dcf7c9e66e21502a58b40cbe627c85 but there was no reason mentioned in the commit for trying to capture all exceptions.

### DIFF
--- a/src/Xmobar/Plugins/StdinReader.hs
+++ b/src/Xmobar/Plugins/StdinReader.hs
@@ -22,16 +22,21 @@ import Prelude
 import System.Posix.Process
 import System.Exit
 import System.IO
-import Control.Exception (SomeException(..), handle)
 import Xmobar.Run.Exec
 import Xmobar.X11.Actions (stripActions)
+import Xmobar.System.Utils (onSomeException)
 
 data StdinReader = StdinReader | UnsafeStdinReader
   deriving (Read, Show)
 
 instance Exec StdinReader where
   start stdinReader cb = do
-    s <- getLine
+    s <-
+      getLine `onSomeException`
+      (\e -> do
+         let errorMessage = "xmobar: Received exception " <> show e
+         hPrint stderr errorMessage
+         cb errorMessage)
     cb $ escape stdinReader s
     eof <- isEOF
     if eof
--- a/src/Xmobar/Plugins/StdinReader.hs
+++ b/src/Xmobar/Plugins/StdinReader.hs
@@ -31,7 +31,7 @@ data StdinReader = StdinReader | UnsafeStdinReader
 
 instance Exec StdinReader where
   start stdinReader cb = do
-    s <- handle (\(SomeException e) -> do hPrint stderr e; return "") getLine
+    s <- getLine
     cb $ escape stdinReader s
     eof <- isEOF
     if eof
--- a/src/Xmobar/Run/Exec.hs
+++ b/src/Xmobar/Run/Exec.hs
@@ -10,7 +10,7 @@
 --
 -- The 'Exec' class and the 'Command' data type.
 --
--- The 'Exec' class rappresents the executable types, whose constructors may
+-- The 'Exec' class represents the executable types, whose constructors may
 -- appear in the 'Config.commands' field of the 'Config.Config' data type.
 --
 -- The 'Command' data type is for OS commands to be run by xmobar
--- a/src/Xmobar/System/Utils.hs
+++ b/src/Xmobar/System/Utils.hs
@@ -17,7 +17,7 @@
 ------------------------------------------------------------------------------
 
 
-module Xmobar.System.Utils (expandHome, changeLoop)
+module Xmobar.System.Utils (expandHome, changeLoop, onSomeException)
 where
 
 import Control.Monad
@@ -25,6 +25,7 @@ import Control.Concurrent.STM
 
 import System.Environment
 import System.FilePath
+import Control.Exception
 
 expandHome :: FilePath -> IO FilePath
 expandHome ('~':'/':path) = fmap (</> path) (getEnv "HOME")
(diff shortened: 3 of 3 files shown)


jaor:
this looks reasonable, but could you please describe what is "the proper behaviour" you can now observe when this exception is risen? also, is there a way of reproducing the error in your environment? (just to play with it myself).

psibi:
@jaor Now the proper behaviour is that there is no CPU throtlling happening. In case of the exception, that thread is killed. And no further update happens to the `StdinReader`. But other part of xmobar works fine (like the date, battery, weather etc)

> also, is there a way of reproducing the error in your environment? (just to play with it myself).

If you are using Ubuntu, just installing xmobar using `nixpkgs` should reproduce it. For fixing it, you would need to set the `LOCALE_ARCHIVE` variable. For triggering the bug, all you need to is enable `StdinReader` plugin and open a browser tab whose tiltle has some unicode content. That will send the the CPU into the busy loop mode. An example URL whichs works for me: https://malayalam.samayam.com/

Another easy of reproducing it(I haven't tried it, but I belive it should work) is to modify StdinReader's code and set it's handle encoding explicitly to `ASCII` and open the above url.

jaor:
@psibi ... i see. one thing that worries me a little is that any transient io error will also deactivate the plugin, which might become responsive again, and perhaps that's what that original PR introducing the handler was trying to address. i'm not sure how much of a concern this should be, i.e., how often (if at all) transient io errors arise (i don't use StdinReader, so i don't have much direct experience with it).  an alternative would be to introduce a wait of a few seconds in the exception handler, so that retries don't clog the CPU, possibly displaying as the value of the monitor the error message, so that users would know immediately that something is going on...

src/Xmobar/Plugins/StdinReader.hs
```diff
@@ -32,8 +32,7 @@ data StdinReader = StdinReader | UnsafeStdinReader
 
 instance Exec StdinReader where
   start stdinReader cb = do
-    s <- handle (\(SomeException e) -> do hPrint stderr e; return "")
-                (hGetLineSafe stdin)
+    s <- hGetLineSafe stdin
     cb $ escape stdinReader s
     eof <- isEOF
```
aij, line 37: Does this still work if EOF errors from hGetLine are no longer handled?

jaor:
On Sun, Apr 26 2020, Ivan Jager wrote:

> @aij commented on this pull request.
>
> ─────────────────────────────────────────────────────────────────────────────────────────────────────────────
>
> In src/Xmobar/Plugins/StdinReader.hs:
>
>> @@ -32,8 +32,7 @@ data StdinReader = StdinReader | UnsafeStdinReader
>
>  instance Exec StdinReader where
>    start stdinReader cb = do
> -    s <- handle (\(SomeException e) -> do hPrint stderr e; return "")
> -                (hGetLineSafe stdin)
> +    s <- hGetLineSafe stdin
>      cb $ escape stdinReader s
>      eof <- isEOF
>
> Does this still work if EOF errors from hGetLine are no longer handled?

that's what i was worrying about in my other comment: not really, it
simply inhabilitates the plugin for good.  if those EOF errors are
expected, we shouldn't remove the handler.

perhaps we could simply return the error string, instead of printing to
stderr, and that way anyone experiencing a permanet problem will see it
in their bar immediately and find the cause sooner.

psibi:
@jaor @aij Thanks for the feedback!

> one thing that worries me a little is that any transient io error will also deactivate the plugin

Are we expecting any specific IO Error that will happen ? Because right now we are capturing all expections (which includes async exceptions) which isn't exactly right.

According to the documentation of `hGetLine`, the operation might fail with an error of `isEOFError`. So probably we are only interested in that ?

> Does this still work if EOF errors from hGetLine are no longer handled?

We get `EOF` errors when the end of file has been reached and if we keep trying read further from the handle inspite of that (which is what we are doing now). Is there any situation where you think we should keep reading inspite of receiving EOF errors ? 

> i see. one thing that worries me a little is that any transient io error will also deactivate the plugin

Is there any other transient IO error you are particularly worried about ? If there is any specific IO exception we are worried about, then we need to handle it specifically instead of catching all exceptions. And for `EOF` errors, is there any particular scenario where you think we should again re-read the handle ?

src/Xmobar/Plugins/StdinReader.hs
psibi: I have addressed your comment here: https://github.com/jaor/xmobar/pull/439#issuecomment-619687393

jaor:
okay, i'm not really sure how often that eof condition happens or how important it is.  we could try and loop only for them.  what i would do in any case is to make sure the error is displayed in the bar, rather than stderr, so that in cases like yours it is easier to discover what's going on.

src/Xmobar/Plugins/StdinReader.hs
```diff
@@ -31,8 +31,7 @@ data StdinReader = StdinReader | UnsafeStdinReader
 
 instance Exec StdinReader where
   start stdinReader cb = do
-    s <- handle (\(SomeException e) -> do hPrint stderr e; return "")
-                (hGetLine stdin)
+    s <- hGetLine stdin
```
jaor, line 34: hlint was telling me that `hGetLine stdin` is just `getLine`

psibi:
@jaor 

> okay, i'm not really sure how often that eof condition happens or how important it is. we could try and loop only for them. 

My point is that we shouldn't try to loop of EOF condition because that error indicates to us that End of file has reached. And we shouldn't read further from the handle after reaching EOF.

> what i would do in any case is to make sure the error is displayed in the bar, rather than stderr, so that in cases like yours it is easier to discover what's going on.

I agree. Displaying the error in the bar would result in much easier way of catching things if something goes wrong. Infact, I think it would be better to log it both on the stderr and the bar. I think I can modify this MR to do these things:
* Catch the IOException
* Update the bar and stderr that we have receivied an error
* Re-throw the exception

Do you agree with the next steps ?

psibi:
@jaor I have gone ahead and modified the behaviour accordingly. The bar will updated properly in case we receive an exception. We also print the error  report on stderr.

jaor:
excellent.  merged, thanks a lot!

PR #441, opened 2020-04-30
Refactor the usage of hGetLineSafe

psibi:
hGetLineSafe is always hGetLine and hence we can directly use it.


jaor:
wow, this was some silly cruft i suppose that left behind in some refactoring.  thanks a lot!

Issue #442, opened 2020-05-04
StdinReader reports EOF error


slotThe:
Starting with commit b7a3d674581720bfb63bf73cae8368ebbad81004 (#439), xmobar---when started from xmonad---now reports an error on the `StdinReader`: 

    xmobar: Received exception <stdin>: hGetLine: end of file

Before this commit, things work perfectly.  This seems to happen with the default xmobar configuration as well.

Minimal xmonad configuration to reproduce this:

```haskell
{-# LANGUAGE NamedFieldPuns #-}

import XMonad
import XMonad.Hooks.DynamicLog (statusBar)

main :: IO ()
main = xmonad
   =<< statusBar "xmobar"
                 def
                 (\XConfig{ modMask } -> (modMask, xK_b))
                 def
```

The `statusBar` function basically takes care of the whole `spawnPipe` shenanigans.  

I'm sort of pressed for time right now, so I don't have much time to investigate this (i.e. starting xmobar manually via `spawnPipe` etc.).  Just wanted to have something to go off of for later debugging.

jaor:
@psibi so it seems this is the scenario the exception handler was trying to prevent.  it seems we need to ignore, at least, EOF exceptions, no?

psibi:
@jaor Probably, but I think we need to understand where the exception is arising from and see if we need to ignore EOF exception.

@slotThe I Just tried the above configuration of yours and it didn't throw exception for me. So, unfortunately I cannot debug this further without help from you. Can you add more details:
* Which platform you are using ?
* What is your xmobar configuration ?

psibi:
@slotThe Also, does the exception occur as soon as you start your xmonad instance or does it take some time for it to receive the exception ?

psibi:
Also, are you able to reproduce the behaviour with the following xmonad configuration:
```
{-# LANGUAGE NamedFieldPuns #-}

import XMonad
import XMonad.Hooks.DynamicLog (statusBar)

main :: IO ()
main = xmonad
   =<< statusBar "xmobar"
                 (def { ppTitle = \_ -> "hello world"})
                 (\XConfig{ modMask } -> (modMask, xK_b))
                 def
```

slotThe:
Thanks for the responses!

>     * Which platform you are using ?

I'm using Void Linux (which does use some non-standard software, like musl libc).  I have, however, also been able to reproduce this on another system running openSUSE Tumbleweed (which uses glibc) with the same xmonad config (and also the one below) and no xmobar config.

[...]

PR #444, opened 2020-05-10
Fix crashes/busy looping happening via fn plugin

psibi:
Right now, with the `StdinReader` plugin enabled - you can crash/cause
busy looping of xmobar if the following html file is opened:

```
<html>
<head>
  <title>hello <fn=1>string</fn> </title>
</head>
</html>
```

More details about this bug is here:
https://github.com/jaor/xmobar/issues/442#issuecomment-625706001

This MR also fixes another bug which produces a crash in xmobar if you
pass non integer items between fn:

<fn=crash>

### DIFF
--- a/src/Xmobar/System/Utils.hs
+++ b/src/Xmobar/System/Utils.hs
@@ -17,11 +17,16 @@
 ------------------------------------------------------------------------------
 
 
-module Xmobar.System.Utils (expandHome, changeLoop, onSomeException)
-where
+module Xmobar.System.Utils
+  ( expandHome
+  , changeLoop
+  , onSomeException
+  , safeIndex
+  ) where
 
 import Control.Monad
 import Control.Concurrent.STM
+import qualified Data.List.NonEmpty as NE
 
 import System.Environment
 import System.FilePath
@@ -50,3 +55,18 @@ onSomeException :: IO a -> (SomeException -> IO b) -> IO a
 onSomeException io what = io `catch` \e -> do _ <- what e
                                               throwIO (e :: SomeException)
 
+(!!?) :: [a] -> Int -> Maybe a
+(!!?) xs i
+    | i < 0     = Nothing
+    | otherwise = go i xs
+  where
+    go :: Int -> [a] -> Maybe a
+    go 0 (x:_)  = Just x
+    go j (_:ys) = go (j - 1) ys
+    go _ []     = Nothing
+{-# INLINE (!!?) #-}
+
+safeIndex :: NE.NonEmpty a -> Int -> a
+safeIndex xs index = case (NE.toList xs) !!? index of
+                       Nothing -> NE.head xs
+                       Just value -> value
--- a/src/Xmobar/App/EventLoop.hs
+++ b/src/Xmobar/App/EventLoop.hs
@@ -40,6 +40,7 @@ import Control.Exception (bracket_, handle, SomeException(..))
 import Data.Bits
 import Data.Map hiding (foldr, map, filter)
 import Data.Maybe (fromJust, isJust)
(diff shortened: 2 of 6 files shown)


src/Xmobar/System/Utils.hs
```diff
@@ -50,3 +55,18 @@ onSomeException :: IO a -> (SomeException -> IO b) -> IO a
...
 
+(!!?) :: [a] -> Int -> Maybe a
+(!!?) xs i
+    | i < 0     = Nothing
+    | otherwise = go i xs
+  where
+    go :: Int -> [a] -> Maybe a
+    go 0 (x:_)  = Just x
+    go j (_:ys) = go (j - 1) ys
+    go _ []     = Nothing
```
jaor, line 66: this definition seems a bit too convoluted.  isn't this just:

```haskell
(??!)  xs i | i < length xs = Just (xs !! i)
           | otherwise = Nothing
```

psibi: @jaor I just took that implementation from the relude library. But your implementation has to traverse the list two times, so it will be slightly inefficient:
* One in the length function
* One via the `!!` function

jaor:
On Sun, May 10 2020, Sibi Prabakaran wrote:

> +(!!?) :: [a] -> Int -> Maybe a
> +(!!?) xs i
> +    | i < 0     = Nothing
> +    | otherwise = go i xs
> +  where
> +    go :: Int -> [a] -> Maybe a
> +    go 0 (x:_)  = Just x
> +    go j (_:ys) = go (j - 1) ys
> +    go _ []     = Nothing
>
> @jaor I just took that implementation from the relude library. But
> your implementation has to traverse the list two times, so it will be
> slightly inefficient:

oh well, O(2n) = O(n), and these lists are tiny anyway.  i'd choose
clarity any day, but if you feel strongly about this implementation and
think it's better, we can keep it.

psibi:
@jaor  I see your point. :-) Since xmobar always runs on my system, I would like it to be as efficient as possible. If you feel strongly, I will replace it with the above simpler implementation.

jaor:
okay then. hlint is giving us some warnings, though:

``` shell
src/Xmobar/System/Utils.hs:70:22: Suggestion: Use fromMaybe
Found:
  case (NE.toList xs) !!? index of
      Nothing -> NE.head xs
      Just value -> value
Perhaps:
  fromMaybe (NE.head xs) ((NE.toList xs) !!? index)

src/Xmobar/System/Utils.hs:70:27: Suggestion: Redundant bracket
Found:
  (NE.toList xs) !!? index
Perhaps:
  NE.toList xs !!? index

src/Xmobar/X11/Parsers.hs:142:32: Warning: Use fromMaybe
Found:
  maybe 0 id
Perhaps:
  Data.Maybe.fromMaybe 0

3 hints
```

(not sure why the last one, i guess it's because my hlint version got updated)

jaor:
running hlint as we do in travis, i.e. with:

``` shell
#!/bin/sh
curl -sSL https://raw.github.com/ndmitchell/neil/master/misc/run.sh | sh -s -- hlint src
```

i only get 2 of those warnings:

``` shell
src/Xmobar/System/Utils.hs:70:27-50: Suggestion: Redundant bracket
Found:
  (NE.toList xs) !!? index
Perhaps:
  NE.toList xs !!? index

src/Xmobar/X11/Parsers.hs:142:32-41: Warning: Use fromMaybe
Found:
  maybe 0 id
Perhaps:
  Data.Maybe.fromMaybe 0

2 hints
```

jaor:
i've fixed the linting myself. merged.

psibi:
Thanks jaor!

PR #448, opened 2020-05-19
Add EOF check before getLine operation from stdin

psibi:
Fixes https://github.com/jaor/xmobar/issues/442

### DIFF
--- a/src/Xmobar/Plugins/StdinReader.hs
+++ b/src/Xmobar/Plugins/StdinReader.hs
@@ -22,21 +22,31 @@ import Prelude
 import System.Posix.Process
 import System.Exit
 import System.IO
-import Control.Exception (SomeException(..), handle)
 import Xmobar.Run.Exec
 import Xmobar.X11.Actions (stripActions)
+import Xmobar.System.Utils (onSomeException)
 
 data StdinReader = StdinReader | UnsafeStdinReader
   deriving (Read, Show)
 
 instance Exec StdinReader where
   start stdinReader cb = do
-    s <- handle (\(SomeException e) -> do hPrint stderr e; return "") getLine
-    cb $ escape stdinReader s
+    -- The EOF check is necessary for certain systems
+    -- More details here https://github.com/jaor/xmobar/issues/442
     eof <- isEOF
     if eof
-      then exitImmediately ExitSuccess
-      else start stdinReader cb
+       then do
+         hPrint stderr $ "xmobar: eof at an early stage"
+         exitImmediately ExitSuccess                
+       else return ()
+    s <-
+      getLine `onSomeException`
+      (\e -> do
+         let errorMessage = "xmobar: Received exception " <> show e
+         hPrint stderr errorMessage
+         cb errorMessage)
+    cb $ escape stdinReader s
+    start stdinReader cb
 
 escape :: StdinReader -> String -> String
 escape StdinReader = stripActions


slotThe:
@jaor I can confirm that this works for me.

Thanks so much to @psibi for being very patient with me while debugging this!

jaor:
merged, together with an hlint nit (using Contro.Monad.when instead of
if).

thank you guys for the relentless pursue :) i'm sorry i couldn't be of
any help, but these weeks i couldn't find the time to switch to an
xmonad environment to try.

liskin:
Did you guys ever figure out why exactly the eof check is needed?

psibi:
@liskin Nah, unfortunately not. My xmobar instance actually worked fine without the eof check. But it seemed to be required for other's machine and @slotThe was incredibly helpful in deterministically reproducing the issue in his machine.

slotThe:
On Thu, Nov 12 2020 12:51, Tomáš Janoušek wrote:
> Did you guys ever figure out why exactly the eof check is needed?

Sadly not; I was looking into this again a while ago and could make
neither heard nor tail of it.

liskin:
:-(

I was really curious about it, as it seems really weird that getLine would throw EOF exception but isEOF wouldn't.
If that really happened that would probably be a bug in GHC runtime or the libc or something.

I tried void linux in a VM and an old xmobar but could reproduce it. But I guess it's pointless to waste any more time on this so I'll need to live without understanding what happened here. :-/

liskin:
And just when I sent this, I think I understood what's going on. :-)

Before fb7279f16d8ccaeab061047b3d30b7d813977bd1, any exception in `getLine` would only be printed and ignored, so the following isEOF would detect an EOF (which @slotThe was getting by restarting xmonad, as xmonad closing the pipe to the old xmobar was the only thing that would tell the old xmobar to exit) and properly exit xmobar. After fb7279f16d8ccaeab061047b3d30b7d813977bd1, the exception isn't handled and it terminates the StdinReader _thread_ without terminating the entire xmobar, causing xmobars to stay there.

The other commit in #439, f3529ad1dbe027e3dda1855a7472aebef6fbe947, added printing of the exception, but it was still rethrown in `onSomeException` causing the thread to fail before getting to call `exitImmediately ExitSuccess`.

So the real problem wasn't the exception being printed (#442). The problem is that not handling the exception terminates the thread but keeps xmobar running, obscuring any new xmobar that is launched when xmonad is restarted.

That also explain why you could only reproduce the issue with lowerOnStart being True: there was some old stuck xmobar all the damn time and no matter what you did with the new one, as long as it lowered itself under the old stuck one, you couldn't see anything else than that.

I'm not really sure how you got the first xmobar stuck, though. I can get it stuck by running xmobar and immediately closing its input, which delivers EOF to the first getLine, but if I ever send any input to it and then close the input pipe, it proceeds to isEOF and terminates correctly. Using `DynamicLog.statusBar` sets up a logHook which does print some stuff to it, so that shouldn't trigger the issue. So this might have been a stray `xmobar` in .xinitrc or .xsession or some `spawn` in xmobar.hs or something like that, something that runs xmonad with stdin connected to /dev/null or an already closed pipe. I can't think of any other explanation. The lowerOnStart mystery really can't be explained by anything else than a second xmobar with no input which refuses to exit due to fb7279f16d8ccaeab061047b3d30b7d813977bd1.

Anyway, the fix you ended up with happens to be correct! Once isEOF returns False, a subsequent getLine should never throw EOF exception, as isEOF blocks until there is some input. It's just that the "The EOF check is necessary for certain systems" comment isn't correct. It's not, as far as I understand it, specific to certain systems, it's just that if the code is reshuffled, it simply doesn't handle early EOF at all, on any system.

Oh and maybe if someone accidentally puts multiple StdinReaders into their xmobar config, the "isEOF returns False, getLine must not fail then" statement isn't true, and this can only be fixed by getting rid of the isEOF check and just properly handling the EOF exception from getLine by exittingImmediately. Not sure if we need to care about such a corner case though.

Please let me know if what I'm saying makes sense. Especially the bit about that stray input-less xmobar needs confirming by @slotThe, as none of his comments have any trace of that xmobar being invoked. :-/

slotThe:
On Fri, Nov 13 2020 16:21, Tomáš Janoušek wrote:
> Please let me know if what I'm saying makes sense. Especially the bit
> about that stray input-less xmobar needs confirming by @slotThe, as
> none of his comments have any trace of that xmobar being invoked. :-/

It makes perfect sense and actually seems to be exactly what's going on!
However, it just leaves me almost more confused than before...

Just to mention this explicitly: there is definitely no stray spawning
of xmobar anywhere on my system.  Just to make sure, I did `rg --hidden
--max-depth 5 xmobar` and went through the output manually; nothing.

Now to the "good" news.  My next (very unscientific) test was to run
`pkill -9 xmobar`, then recompile via M-q and just list the relevant
processes.

ea1096f20031364b5cffb0eebd64a6dc1a6e4a6e (current HEAD, works):

    slot     21154  0.0  0.0   1004     4 ?        Ss   08:28   0:00 /bin/sh -c xmobar
    slot     21155  2.5  0.4 1073779900 38540 ?    S    08:28   0:00 xmobar
    slot     21160  0.0  0.0   1748   520 pts/7    S+   08:28   0:00 grep xmobar

b7a3d674581720bfb63bf73cae8368ebbad81004 (broken commit):

    slot     16483  0.0  0.0   1004     4 ?        Ss   08:22   0:00 /bin/sh -c xmobar
    slot     16485  0.1  0.1 1073780824 10764 ?    S    08:22   0:00 xmobar
    slot     16504  0.0  0.0   1004     4 ?        Ss   08:22   0:00 /bin/sh -c xmobar
    slot     16505  0.1  0.1 1073780824 10588 ?    S    08:22   0:00 xmobar
    slot     16506  0.0  0.0   1004     4 ?        Ss   08:22   0:00 /bin/sh -c xmobar
    slot     16507  0.0  0.1 1073780824 10776 ?    S    08:22   0:00 xmobar
    slot     16530  0.0  0.0   1004     4 ?        Ss   08:22   0:00 /bin/sh -c xmobar
    slot     16531  0.1  0.1 1073780868 10356 ?    S    08:22   0:00 xmobar
    slot     16701  0.0  0.0   1748   556 pts/5    S+   08:23   0:00 grep xmobar

Bingo!

Now for the part that I don't understand.  The `/bin/sh -c xmobar` are
processes that are spawned by `spawnPipe`; which means that, for some
reason, `spawnPipe` is acting out and spawns multiple xmobars.  Going
through my logs I can see that this is still happening (i.e. I still get
"xmobar: eof at an early stage" when I restart xmonad), just now these
xmobars are properly terminated and so it appears to be fine.

liskin:
Okay, @slotThe, thanks for confirming that the issue is indeed caused by lingering xmobars. That means the comment in StdinReader is misleading, and as I found another way to cause an xmobar to linger, I'm submitting this fix/refactor: https://github.com/jaor/xmobar/pull/505

I haven't looked yet into why your xmonad launches several xmobars which get immediately EOFd, but I intend to study this further in the afternoon. If this is a bug in xmonad(-contrib), we definitely want to get it fixed!

liskin:
@slotThe Here's what's going on:

your minimal config (and also unfortunately the config recommended by X.H.DynamicLog) says
```
main = xmonad =<< statusBar "xmobar" def (\XConfig{ modMask } -> (modMask, xK_b)) def
```
and `statusBar` immediately spawnsPipe xmobar (instead of doing that in the startupHook), so when you M-q to **recompile** and restart xmonad, it spawns a useless xmobar just for the recompilation step. Now I have no idea why in your case there are three extra xmobars, that is a bit weird, but I guess if we fix this there'll be just one.

I'll patch DynamicLog to spawn the xmobar in startupHook and I'll give you a PR for testing.

liskin:
DynamicLog pull request here: https://github.com/xmonad/xmonad-contrib/pull/400
Can you please test if this brings the number of extra xmobars down?

PR #450, opened 2020-05-30
Optimize weather plugin by reusing manager and other refactors

psibi:
As documented in the http-client library, calling newManager is an
expensive operation:

```
Creating a new Manager is a relatively expensive operation, you are
advised to share a single Manager between requests instead.
```

But inspite of the haddocks in xmobar claiming that once 'Manager' is
created, it will be used throughout the monitor is not true. Because for
every call of `startWeather` a new manager is being created.

Also I removed the option in WeatherOpts because even if it is false,
it will be ultimately created in `getData` function. Also without
using a manager - the plugin won't really work. So, I don't think
there is any reason for this option to exist.

I have introduced a new dependency http-client-tls to use the shared
global manager so that we reuse the same manager every time. This
simplifies a lot of code. Note that this is not really a new
dependency because http-conduit already depends on it transitively.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,8 @@ _New features_
   - Optimize date plugin by avoiding calling getTimeZone for each of
     the time the date has to be updated. Instead, it's computed once
     at the start and re-used for each invocation.
+  - Optimize Weather and UVMeter plugin by using global Manager instead of
+    creating for each http request when useManager is explicitly configured as False.
 
 ## Version 0.33 (February, 2020)
 
--- a/src/Xmobar/Plugins/Monitors/UVMeter.hs
+++ b/src/Xmobar/Plugins/Monitors/UVMeter.hs
@@ -21,14 +21,11 @@ import qualified Control.Exception as CE
 import Network.HTTP.Conduit
     ( Manager
     , httpLbs
-    , managerConnCount
-    , newManager
     , parseRequest
     , responseBody
-    , tlsManagerSettings
     )
+import Network.HTTP.Client.TLS (getGlobalManager)
 import Data.ByteString.Lazy.Char8 as B
-import Data.Maybe (fromMaybe)
 import System.Console.GetOpt (ArgDescr(ReqArg), OptDescr(Option))
 import Text.Read (readMaybe)
 import Text.Parsec
@@ -66,12 +63,9 @@ uvURL :: String
 uvURL = "https://uvdata.arpansa.gov.au/xml/uvvalues.xml"
 
 -- | Get the UV data from the given url.
-getData :: Maybe Manager -> IO String
-getData uvMan = CE.catch
-    (do man <- flip fromMaybe uvMan <$> mkManager
-        -- Create a new manager if none was present or the user does not want to
-        -- use one, otherwise use the provided manager.
-        request <- parseRequest uvURL
+getData ::Manager -> IO String
+getData man = CE.catch
+    (do request <- parseRequest uvURL
         res <- httpLbs request man
         return $ B.unpack $ responseBody res)
     errHandler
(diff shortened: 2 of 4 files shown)


jaor:
hmm, this doesn't seem to work for me, my monitor stays in "Updating..." when i start xmobar

jaor:
ah, wait, i had the -m option set in my config...

jaor:
okay, it's working now.  this looks good. we had some issues when @slotThe implemented this because for some unknown reasons using a manager caused (my) xmobar CPU consumption to increase, and we decided to introduce the -m flag... but i'm not seeing this problem now, and i like the simpler code.  @slotThe do you see any problem? also, UVMeter uses a manager and should be modifier in the same way.

slotThe:
On Sat, May 30 2020 11:11, jao wrote: 
> @slotThe do you see any problem? 

This looks good to me.  As you said, UVMeter should be modified in 
the same way, but other than that I do like the cleaner code as 
well!

> But inspite of the haddocks in xmobar claiming that once 
> 'Manager' is created, it will be used throughout the monitor is 
> not true. Because for every call of startWeather a new manager 
> is being created.

This is a bit puzzling to me, however.  The way I read the code,
startWeather gets called exactly once (when xmobar first starts up).
After that (i.e. when the monitor is updated) everything is handled by
running the underlying reader, waiting, and then repeating that process
(in runMLD), which only calls runWeather (because it's the action we
explicitly give to the runMD function in startWeather').  What am I
missing here?

psibi:
> This is a bit puzzling to me, however.  The way I read the code,
startWeather gets called exactly once (when xmobar first starts up).
After that (i.e. when the monitor is updated) everything is handled by
running the underlying reader, waiting, and then repeating that process
(in runMLD), which only calls runWeather (because it's the action we
explicitly give to the runMD function in startWeather').  

You're right! The whole recreation thing only happens when the `useManager` is explicitly configured as `False`. For the default scenario, that isn't the case.

BTW, I have created three new issues related to optimization. I would like to hear your views on that if possible.

jaor:
@psibi do you think we could add UVMeter to this PR? also, hlint is telling me this:

```
src/Xmobar/Plugins/Monitors/Weather.hs:(34,1)-(36,3): Suggestion: Use newtype instead of data
Found:
  data WeatherOpts = WeatherOpts {weatherString :: String}
Perhaps:
  newtype WeatherOpts = WeatherOpts {weatherString :: String}
Note: decreases laziness
```

psibi:
@jaor Sure, I could do that by the end of this week. I think we need hlint integrated with this repo. :-)

jaor:
On Wed, Jun 03 2020, Sibi Prabakaran wrote:

> @jaor Sure, I could do that by the end of this week. I think we need
> hlint integrated with this repo. :-)

it's already integrated in the travis tests that are run after a commit
to master.  i think it can be configured per PR too.

jaor:
as a matter of fact:

https://travis-ci.org/github/jaor/xmobar/builds/692898508

psibi:
@jaor Ah, I missed that. Do you know why the builds aren't getting shown in the MR ?

psibi:
@jaor 

>  i've made that so, but they're still not
showing: maybe it'll pop up when something is pushed or a new PR is
created...

I had a similar issue in another open source repository which I was contributing to recently. They weren't able to figure out the reason and they just migrated to github's CI. Might be an issue with Travis CI.

psibi:
@jaor I will finish this MR as part of this week's Zurihac hackathon. I have added that project to this year's Zurihac as I'm planning to work on xmobar: https://github.com/zfoh/zfoh.ch/pull/110

If you would have some time available, it would be nice to have you too (and others in the community: CC: @slotThe )

jaor:
That's great, Sibi. I'd really like to participate, but I'm not sure right now if I am going to be free. Something I would have liked to try is to move away from Xft, on to Cairo... text placement and alignment right now is flimsy to say the least, and I've never been able to make real sense of the Xft api. Some form of "system tray" (or its evolution in recent times) support used also to be high in my list...

-------- Original Message --------

psibi:
@jaor Thanks for the update!

I have pushed some commits which updates the MR. :-)

jaor:
excellent. merged!

Issue #451, opened 2020-06-01
Optimizing startCommand


psibi:
Right now the startCommand has the following code:
```
    | otherwise = do var <- atomically $ newTVar is
                     let cb str = atomically $ writeTVar var (s ++ str ++ ss)
                     a1 <- async $ start com cb
                     a2 <- async $ trigger com $ maybe (return ())
                                                 (atomically . putTMVar sig)
```

We seem to be spawning two green threads for each of the commands. Looking into the code, I don't really see why `a2` is needed. Infact, I have been running xmobar for the past day without it and everything seems to work fine for me. Any specific reason why it's needed ?

PR #455, opened 2020-06-05
Optimize startCommand

psibi:
For every monitor in xmobar, we currently spawn two threads during
startup. But one of the threads from it get's exited as soon as it's
started because this is the default definition of trigger:

```
    trigger :: e -> (Maybe SignalType -> IO ()) -> IO ()
    trigger _ sh  = sh Nothing
```

So, for the `Nothing`, the expression `maybe (return ()) (atomically
. putTMVar sig)` evaluates to `()` and returns immediately. The only
place where this seems to be required is the `BufferedPipeReader`
plugin. I'm moving the spawing of the green thread to that plugin
itself so as to avoid this overhead for the other monitors where this
isn't used.

Fixes https://github.com/jaor/xmobar/issues/451

### DIFF
--- a/src/Xmobar/Plugins/BufferedPipeReader.hs
+++ b/src/Xmobar/Plugins/BufferedPipeReader.hs
@@ -16,6 +16,7 @@ module Xmobar.Plugins.BufferedPipeReader(BufferedPipeReader(..)) where
 
 import Control.Monad(forM_, when, void)
 import Control.Concurrent
+import Control.Concurrent.Async (withAsync, wait)
 import Control.Concurrent.STM
 import System.IO
 import System.IO.Unsafe(unsafePerformIO)
@@ -35,7 +36,9 @@ instance Exec BufferedPipeReader where
     alias      ( BufferedPipeReader a _  )    = a
 
     trigger br@( BufferedPipeReader _ _  ) sh =
-        takeMVar signal >>= sh . Just >> trigger br sh
+        let action = newTrigger br sh
+            newTrigger br2 sh2 = takeMVar signal >>= sh2 . Just >> newTrigger br2 sh2
+        in withAsync action wait
 
     start      ( BufferedPipeReader _ ps ) cb = do
 
--- a/src/Xmobar/App/EventLoop.hs
+++ b/src/Xmobar/App/EventLoop.hs
@@ -245,9 +245,8 @@ startCommand sig (com,s,ss)
     | otherwise = do var <- atomically $ newTVar is
                      let cb str = atomically $ writeTVar var (s ++ str ++ ss)
                      a1 <- async $ start com cb
-                     a2 <- async $ trigger com $ maybe (return ())
-                                                 (atomically . putTMVar sig)
-                     return ([a1, a2], var)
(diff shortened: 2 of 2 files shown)


psibi:
>  and remove an abstraction that nobody uses. What do you think?

I like your idea. The only thing we need to figure out is how to pass the signal type which we are passing to the trigger function. I don't want to add an extra argument to the `start` function to achieve this. I will think more about this to see if we can achieve this by removing the `trigger` function without introducing any new parameters to the `start` function.

psibi:
@jaor I looked into this and it seems that I would need to introduce one extra parameter to the `start` function to achieve it - if we plan to remove the whole `trigger` machinery. What do you think ?

I'm leaning towards the current design because of it as it's a more simple change. :-)

jaor:
On Fri, Jun 12 2020, Sibi Prabakaran wrote:

> @jaor I looked into this and it seems that I would need to introduce
> one extra parameter to the start function to achieve it - if we plan
> to remove the whole trigger machinery. What do you think ?
>
> I'm leaning towards the current design because of it as it's a more
> simple change. :-)

yes, i think i agree... complicating start for every monitor seems worse
than keeping the mostly ignorable trigger!

src/Xmobar/Plugins/BufferedPipeReader.hs
```diff
@@ -35,7 +36,9 @@ instance Exec BufferedPipeReader where
     alias      ( BufferedPipeReader a _  )    = a
 
     trigger br@( BufferedPipeReader _ _  ) sh =
-        takeMVar signal >>= sh . Just >> trigger br sh
+        let action = newTrigger br sh
+            newTrigger br2 sh2 = takeMVar signal >>= sh2 . Just >> newTrigger br2 sh2
```
liskin, line 40: There's no need to pass `br2` and `sh2` here, what about:

```
let go = takeMVar signal >>= sh . Just >> go
in async go
```

src/Xmobar/Plugins/BufferedPipeReader.hs
```diff
@@ -35,7 +36,9 @@ instance Exec BufferedPipeReader where
...
+        in withAsync action wait
```
liskin, line 41: You sure about the `wait`? By dropping `async $ …` from `startCommand` you changed the semantics of `trigger` from a "trigger thread loop" to "start the trigger thread", so this needs to not be blocking, and `withAsync`/`wait` shouldn't be used here.

In fact you probably want to just move the `async $ …` from `startCommand` here and change the return type of `trigger accordingly so that the async handle of the trigger thread (if any, so wrapped in `Maybe`) can still be returned from `startCommand`.

psibi:
@liskin Thanks for your feedback! Since this optimisation is a one-time cost and doesn't improve perf once the thread exits - I'm not keen to pursue it further. Others can pick it up if they are interested. I'm closing the MR.

liskin:
@psibi Okay. :)

I vaguely remember that last time I touched `Exec`, `trigger` triggered me and I wanted to do something with it, but I can't for the life of me remember what it was. Probably wasn't all that important, like this. Let's not worry about it then. :-)))

PR #456, opened 2020-06-05
Optimize date plugin

psibi:
We avoid calling getTimeZone for each of the time the date has to be
updated. Instead, it's computed once at the start and re-used for each
invocation.

Looking at the implementation of 'getTimeZone', we can see that it's
very expensive:
https://www.stackage.org/haddock/lts-15.15/time-1.9.3/src/Data-Time-LocalTime-Internal-TimeZone.html#getTimeZone

It calls a C FFI each time to get the time
zone (getTimeZoneCTime). This is something which we can avoid and the
MR implements that.

I have been using my xmobar with this patch and the result has been
quite good. My xmobar CPU usage has used to hit 3~7%
intermittently. With this MR, It hits only 0.7% intermittently which
is nice. :-)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,9 @@ _New features_
   - New plugin `HandleReader` for reading data from a Haskell `Handle`. This is
     useful if you are running xmobar from within a Haskell program.
   - Build with ghc 8.10 allowed.
+  - Optimize date plugin by avoiding calling getTimeZone for each of
+    the time the date has to be updated. Instead, it's computed once
+    at the start and re-used for each invocation.
 
 ## Version 0.33 (February, 2020)
 
--- a/src/Xmobar/Plugins/Date.hs
+++ b/src/Xmobar/Plugins/Date.hs
@@ -31,8 +31,16 @@ data Date = Date String String Int
 
 instance Exec Date where
     alias (Date _ a _) = a
-    run   (Date f _ _) = date f
     rate  (Date _ _ r) = r
+    start (Date f _ r) cb = do
+      t <- getCurrentTime
+      zone <- getTimeZone t
+      go zone
+     where
+      go zone = doEveryTenthSeconds r $ date zone f >>= cb
 
-date :: String -> IO String
-date format = fmap (formatTime defaultTimeLocale format) getZonedTime
+date :: TimeZone -> String -> IO String
+date timezone format = do
+  time <- getCurrentTime
+  let zonedTime = utcToZonedTime timezone time
+  pure $ formatTime defaultTimeLocale format zonedTime


jaor:
excellent! merged

liskin:
Hey @psibi are you sure the improvement from several percent to 0.7 is from this patch? I just pulled this and can't see any performance improvement at all, so I'm wondering if perhaps you forgot to fetch the master (https://github.com/jaor/xmobar/pull/410 and https://github.com/jaor/xmobar/pull/409) before starting to benchmark this?

I was expecting that this would result in my xmobar going from 0.3-0.7 to, say, 0.1 overall, but it didn't and the CPU usage pattern seems exactly the same, so I'll be probably reverting this locally so I don't need to worry about restarting stuff when travelling.

psibi:
@liskin For my testing, I only used the `Date` plugin in my configuration to observe it. But observing it manually is not totally scientific and easy to get it wrong. That's why I have introduced the benchmark support in xmobar recently: https://github.com/jaor/xmobar/pull/458

In the new implementation we are avoiding one FFI call for each invocation. So that should definitely add some improvements, although it would be nice to get it confirmed from the `gauge` benchmarks.

liskin:
Interesting... I wonder where those several percents came from. It's also possible that my system simply has cheaper getTimeZone for some completely unrelated reasons. :-)

Don't worry about it then, I don't mind keeping local patches.

psibi:
@liskin 

> It's also possible that my system simply has cheaper getTimeZone for some completely unrelated reasons. :-)

What system are you using ? :-) There were two code paths for getTimeZone IIRC (Windows and Linux). But FFI calls in general are expensive and it's best if we can avoid it if possible.

liskin:
amd64 Debian testing. I'd expect FFI calls to be considerably less expensive than the syscalls xmobar performs to read all those CPU and network and battery stats, but that's just a guess. :-)

psibi:
@liskin 

> so I'm wondering if perhaps you forgot to fetch the master (#410 and #409) before starting to benchmark this?

Note that I did have them in the master before bench-marking this. Also, I'm still not satisfied with xmobar's perf as of now. I see that that the main thread of `xmobar` still has very intermittent high CPU usage (I'm guessing this happens whenever the main co-ordination thread becomes active). It still constantly jumps to 5-10% intermittently for me whenever I enable all the plugins. I was wondering how  you measured/profile xmobar when you implemented the timer coalescing ? I looked on powertop but there seems be no proper way to track the wakeups per process.

liskin:
Nothing fancy. I just looked at top, at the battery watts xmobar itself reports, and used GHC time profiling and perf. Using perf I found out that reading the battery status takes several milliseconds, so decreased the refresh frequency of that. GHC profiling helped me optimize the Net monitor. Timer coalescing was then mostly driven by me looking at strace and having an idea what I want the output to look like. :-)

liskin:
Oh and re high CPU usage of the main thread: don't forget it's that thread which does the rendering. Might consume more CPU if you use Xft (I don't).

jaor:
On Tue, Jun 16 2020, Tomáš Janoušek wrote:

> so I'll be probably reverting this locally so I don't need to worry
> about restarting stuff when travelling.

hmm, i hadn't realized this change actually modified the plugin's
behaviour.  for me, having to restart xmobar because i moved the
computer is also a no-go... does this happen, e.g., when changing the
network to where the computer is connected? (i see it working well with
sleep).

for me, moreover, this plugin updates every 30 minutes, so any
performance gains are totally lost :)

liskin:
> does this happen, e.g., when changing the network to where the computer is connected?

Well if you have some daemon which changes timezone depending on what network you connect to then yeah. I'm currently switching manually but I intend to use something based on geoclue when I get to it.

psibi:
@liskin 

> Oh and re high CPU usage of the main thread: don't forget it's that thread which does the rendering. Might consume more CPU if you use Xft (I don't).

That's a valid point! I use `Xft` so probably that's causing the CPU issue for me still. I was suspecting if the time API's being used in the timer loop is causing the CPU usage, but `Xft` is a good suspect. I will try debug this further. Thanks!

> Timer coalescing was then mostly driven by me looking at strace and having an idea what I want the output to look like. :-)

Ah, okay. That's good to know. I thought you were using `powertop` to optimize it. But using `powertop` isn't nice in my experience.

@jaor 

> for me, moreover, this plugin updates every 30 minutes, so any performance gains are totally lost :)

I run it every minute, so it's more important for me. :-)

liskin:
> I run it every minute, so it's more important for me. :-)

My xmobar updates the date/time every second. Without this patch, strace shows stat("/etc/localtime", …) every second. But it's just one syscall out of many (my xmobar reads 5 /proc|/sys files every second and 21 more every 5 seconds), which explains why I'm not seeing any performance improvement.

jaor:
On Tue, Jun 16 2020, Sibi Prabakaran wrote:

>     for me, moreover, this plugin updates every 30 minutes, so any
>     performance gains are totally lost :)
>    
> I run it every minute, so it's more important for me. :-)


oops sorry, i was thinking all the time of the weather plugin when
writing those comments! of course, my date plugin also runs every minute
(and having to restart it if one changes timezone perhaps less critical,
althought it'd be nice to avoid).  i haven't seen a big perf difference,
but then i use the TopMem and TopProc monitors, which are dominating the
cycles of my xmobar most of the time.

psibi:
> i haven't seen a big perf difference, but then i use the TopMem and TopProc monitors, which are dominating the cycles of my xmobar most of the time.

I did use `TopProc` for a couple of days. But unfortunately it seem to be little extra CPU intensive in my machine compared to the other monitors, that I had to drop using it. Although I would love to have it in my bar some day. :-)

PR #458, opened 2020-06-13
Initial support for benchmarks of the plugins

psibi:
This MR adds the initial infrastructure for the benchmarking of the plugins. Right now it's quite hard to measure things and see if something is actually causing any improvements here.

This is the output you get right now with this MR:
```
$ stack bench
benchmarked Cpu Benchmarks/CPU normal args
time                 100.5 μs   (99.87 μs .. 100.8 μs)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 102.2 μs   (101.7 μs .. 103.4 μs)
std dev              2.377 μs   (1.009 μs .. 4.864 μs)

Benchmark xmobarbench: FINISH
Completed 2 action(s).
```

Note that this is initial MR and we should improve this further by adding benchmark for various other monitors.

### DIFF
--- a/bench/main.hs
+++ b/bench/main.hs
@@ -1,0 +1,44 @@
+{-#LANGUAGE RecordWildCards#-}
+
+import Gauge
+import Xmobar
+import Xmobar.Plugins.Monitors.Common.Types
+import Xmobar.Plugins.Monitors.Common.Run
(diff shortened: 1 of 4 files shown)


jaor:
As promising as sorely needed. Thanks!

PR #460, opened 2020-06-14
Optimize CPU Monitor: 25% perf boosts and other changes

psibi:
Summary of the changes:

* Make the CPU monitor more type safe. Added various new types instead of having `[String]` and `[Float]` everywhere.
* Add test cases for the CPU monitor
* Introduce new type named `PureConfig`. This is a pure version of `MConfig` and the initial work has been done in this MR which will hopefully in the future remove the need of the whole `MConfig` type.
* Avoid template parsing for the components which won't be rendered
* Avoid formatting output for the components which won't be rendered
* Perf improvements:

Before:
```
benchmarked Cpu Benchmarks/CPU normal args
time                 101.6 μs   (100.8 μs .. 102.3 μs)
                     0.999 R²   (0.997 R² .. 1.000 R²)
mean                 103.0 μs   (102.4 μs .. 104.5 μs)
std dev              2.883 μs   (1.465 μs .. 5.154 μs)
variance introduced by outliers: 11% (moderately inflated)
```
After:
```
time                 75.80 μs   (75.15 μs .. 76.36 μs)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 75.97 μs   (75.74 μs .. 76.41 μs)
std dev              1.057 μs   (666.6 ns .. 1.927 μs)
```

### DIFF
--- a/src/Xmobar/Plugins/Monitors/Common/Output.hs
+++ b/src/Xmobar/Plugins/Monitors/Common/Output.hs
@@ -101,8 +101,8 @@ pSetColor config str s =
             Just c -> "<fc=" ++ c ++ ">" ++ str ++ "</fc>"
 
 pShowWithPadding :: MonitorConfig -> String -> String
-pShowWithPadding MonitorConfig {..} s =
-  padString pMinWidth pMaxWidth pPadChars pPadRight pMaxWidthEllipsis s
+pShowWithPadding MonitorConfig {..} =
+  padString pMinWidth pMaxWidth pPadChars pPadRight pMaxWidthEllipsis
 
 pFloatToPercent :: MonitorConfig -> Float -> String
 pFloatToPercent MonitorConfig{..} n = let p = showDigits 0 (n * 100)
--- a/src/Xmobar/Plugins/Monitors/Common/Output.hs
+++ b/src/Xmobar/Plugins/Monitors/Common/Output.hs
@@ -43,6 +43,7 @@ module Xmobar.Plugins.Monitors.Common.Output ( IconPattern
                                              , pShowPercentBar
                                              , pShowVerticalBar
                                              , pShowIconPattern
+                                             , pShowPercentWithColors
                                              ) where
 
 import Data.Char
--- a/src/Xmobar/Plugins/Monitors/Common/Output.hs
+++ b/src/Xmobar/Plugins/Monitors/Common/Output.hs
@@ -56,7 +56,7 @@ import Xmobar.Plugins.Monitors.Common.Types
 
 type IconPattern = Int -> String
 
-pShowVerticalBar :: (MonadIO m) => PureConfig -> Float -> Float -> m String
+pShowVerticalBar :: (MonadIO m) => MonitorConfig -> Float -> Float -> m String
 pShowVerticalBar p v x = pColorizeString p v [convert $ 100 * x]
   where convert :: Float -> Char
         convert val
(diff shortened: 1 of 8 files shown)


jaor:
This looks very good indeed.  I am very happy to see tests too.  I have just a nit, that might be a bit annoying, but here it goes: i don't like the `PureConfig` name, because 'pure' just describes an implementation detail, which is neither here nor there (it feels like "typed Hungarian notation" if you take my meaning).  How about we call it simply `MonitorConfig`?

psibi:
@jaor Thanks for the feedback! I will do the changes tomorrow. Free feel to let me know other changes too (including naming!). :-)

aniketd:
Awesome! Thanks @psibi !

src/Xmobar/Plugins/Monitors/Common/Output.hs
```diff
@@ -37,18 +39,89 @@ module Xmobar.Plugins.Monitors.Common.Output ( IconPattern
...
 type IconPattern = Int -> String
 
+pShowVerticalBar :: (MonadIO m) => PureConfig -> Float -> Float -> m String
+pShowVerticalBar p v x = pColorizeString p v [convert $ 100 * x]
+  where convert :: Float -> Char
+        convert val
+          | t <= 9600 = ' '
+          | t > 9608 = chr 9608
+          | otherwise = chr t
+          where t = 9600 + (round val `div` 12)
```
slotThe, line 66: Might be worth adding a comment here what these magic numbers do

src/Xmobar/Plugins/Monitors/Common/Output.hs
```diff
@@ -37,18 +39,89 @@ module Xmobar.Plugins.Monitors.Common.Output ( IconPattern
...
+     zipWithM (pShowWithColors p . const) fstrs temp
+
+pShowPercentWithColors :: (MonadIO m) => PureConfig -> Float -> m String
+pShowPercentWithColors p f = fmap head $ pShowPercentsWithColors p [f]
+
+pShowPercentBar :: (MonadIO m) => PureConfig -> Float -> Float -> m String
+pShowPercentBar p@PureConfig{..} v x = do
+  let bb = pBarBack
+      bf = pBarFore
+      bw = pBarWidth
```
slotThe, line 81: All of the uses of `RecordWildCards` I've seen look a little bit like this. If you're doing this kind of renaming anyways, how about using `NamedFieldPuns` instead of `RecordWildCards`?

src/Xmobar/Plugins/Monitors/Common/Output.hs
```diff
@@ -37,18 +39,89 @@ module Xmobar.Plugins.Monitors.Common.Output ( IconPattern
...
+
+pShowWithColors :: (Num a, Ord a, MonadIO m) => PureConfig -> (a -> String) -> a -> m String
+pShowWithColors p f x = do
+  let str = pShowWithPadding p (f x)
+  pColorizeString p x str
+
+pColorizeString :: (Num a, Ord a, MonadIO m) => PureConfig -> a -> String -> m String
+pColorizeString p x s = do
+    let h = pHigh p
+        l = pLow p
```
slotThe, line 94: I guess "for consistency" this should either use `RecordWildCards` or `NamedFieldPuns` as well

src/Xmobar/Plugins/Monitors/Common/Output.hs
```diff
@@ -174,6 +247,15 @@ showIconPattern (Just str) x = return $ str $ convert $ 100 * x
...
 
+pShowIconPattern :: Maybe IconPattern -> Float -> IO String
+pShowIconPattern Nothing _ = return ""
+pShowIconPattern (Just str) x = return $ str $ convert $ 100 * x
+  where convert val
+          | t <= 0 = 0
+          | t > 8 = 8
+          | otherwise = t
+          where t = round val `div` 12
+
```
slotThe, line 258: Maybe also add a comment what these magic numbers mean

src/Xmobar/Plugins/Monitors/Common/Output.hs
psibi: @slotThe These are direct adaptation of code from the monad transformer version of the code. I merely ported them to a non monad transformer version here.

src/Xmobar/Plugins/Monitors/Common/Output.hs
psibi: That's an oversight and I have to remove the usage of `bb`, `bf` etc. I will do that. Thanks!

src/Xmobar/Plugins/Monitors/Common/Output.hs
psibi: Same comment as here: https://github.com/jaor/xmobar/pull/460#discussion_r441409251 Thanks for pointing out.

src/Xmobar/Plugins/Monitors/Common/Output.hs
psibi: Same comment as here: https://github.com/jaor/xmobar/pull/460#discussion_r441408898

psibi:
@jaor / @slotThe I have updated the PR to address your comments.

jaor:
merged, thanks! (even travis-ci is working now!)

PR #477, opened 2020-07-24
Add a new tag (<box>) to set borders around text (can be used to underline)

### LINKED ISSUE #238: Text background color
Hello,
I have problem with setting background color behind some text (with <fc> tag). It seems, that background fills area with height of text and that height is somehow wrong. I don't know whether it is issue with my font or with xmobar calculating its size. I thought that filling whole height of bar would fix it, but then I realized, that those Powerline-like arrows (which should be with maximal font height) would never connect with that filled background correctly. I have even problems with automatical vertical centering - text is shifted to the top of the bar and not centered. That should be easy to cope with with setting textOffset to some value. But that only makes background issue even worse.

Screenshot of how that looks is here:
http://imgur.com/fWYgmpx

My xmobar configuration:
http://pastebin.com/PNk0cA7m

Just ignore strange characters in configuration, I am tweaking my fonts...

Unoqwy:
- The `<fc>` tag can take two new parameters (separated by `:` from colors): top offset and bottom offset for the background. This is intended to fix #238. Usage: `<fc=white,black:5>`, `<fc=red:10,10`>, etc
- ColorString has been edited and refactored to TextRenderInfo as it now holds more things.
- The `<box>` tag can be used to underline, add side borders, top borders, etc. See examples below

`<box>` examples:
`<box=BBFull L 0 1 white>foo</box>` (`<box=BBFull>` will also default to this) creates a full white border with 1 line width
`<box=BBBottom C 20 3 red>bar</box>` underline the text with 20 offset on each side (because of C) in red

`<box>`usage: `<box=Position [Align [Offset [Width [Color]]]]`
default values: Align = C, Offset = 0, Width = 1, Color = white

Position values: `BBTop`, `BBBottom`, `BBVBoth`, `BBLeft`, `BBRight`, `BBHBoth`, `BBFull`
Align values: `L`, `C`, `R`

Note: I started Haskell-ing only 3 days ago, I may have used some bad practices (I hope not tho). Any feedback is well appreciated :+1: 

Example of usage:


jaor:
This looks generally good.  There are a few hlint warnings to fix: you
can get them with:

``` shell
$ curl -sSL https://raw.github.com/ndmitchell/neil/master/misc/run.sh | sh -s -- hlint src
```

(copying the results below)

It'd be also very nice if you could please update the documentation in
readme.md with the new options.

Thanks a lot!


``` shell

src/Xmobar/X11/Draw.hs:156:32-48: Suggestion: Redundant bracket
Found:
  (fromIntegral ht) - ot
Perhaps:
  fromIntegral ht - ot

src/Xmobar/X11/Draw.hs:163:37-58: Warning: Use notElem
Found:
  not (b `elem` triBoxes)
Perhaps:
  notElem b triBoxes

src/Xmobar/X11/Draw.hs:165:21-51: Suggestion: Use tuple-section
Found:
  \ b -> ((offset, offset + l), b)
Perhaps:
  ((offset, offset + l),)
Note: may require `{-# LANGUAGE TupleSections #-}` adding to the top of the file

src/Xmobar/X11/Draw.hs:165:67-81: Suggestion: Redundant bracket
Found:
  triBoxes \\ (map snd boxes)
Perhaps:
  triBoxes \\ map snd boxes

src/Xmobar/X11/Draw.hs:196:32-40: Suggestion: Redundant bracket
Found:
  (0, (- offset))
Perhaps:
  (0, - offset)

src/Xmobar/X11/Draw.hs:197:32-40: Suggestion: Redundant bracket
Found:
  (offset, (- offset))
Perhaps:
  (offset, - offset)

src/Xmobar/X11/Parsers.hs:43:18-30: Suggestion: Redundant bracket
Found:
  case (words input) of
    [pos] -> Just $ Box (read pos) C 0 1 "white"
    [pos, alg] -> Just $ Box (read pos) (read alg) 0 1 "white"
    [pos, alg, off]
      -> Just $ Box (read pos) (read alg) (read off) 1 "white"
    [pos, alg, off, wdh]
      -> Just $ Box (read pos) (read alg) (read off) (read wdh) "white"
    [pos, alg, off, wdh, c]
      -> Just $ Box (read pos) (read alg) (read off) (read wdh) c
    _ -> Nothing
Perhaps:
  case words input of
    [pos] -> Just $ Box (read pos) C 0 1 "white"
    [pos, alg] -> Just $ Box (read pos) (read alg) 0 1 "white"
    [pos, alg, off]
      -> Just $ Box (read pos) (read alg) (read off) 1 "white"
    [pos, alg, off, wdh]
      -> Just $ Box (read pos) (read alg) (read off) (read wdh) "white"
    [pos, alg, off, wdh, c]
      -> Just $ Box (read pos) (read alg) (read off) (read wdh) c
    _ -> Nothing

7 hints
```

jaor:
a question: i don't understand how specifying both a bottom and top text offset would work... seems that given the font size, only one of them can be given.  it's also a bit strange looking that fc (font color) takes this new parameter (we use separated ones in other places, e.g. fn)... it'd be better if we could have a new tag, say `fp`.  what do you think?

Unoqwy:
I'm a bit busy today, I'll fix the hints and update the readme tomorrow.
The offset parameters on `<fc>` are meant for the background. Say for example we don't use the offset parameter, the background of the text would be the same height as the rendered text which in some cases, we don't want. It allows the background color to fill the bar the way we want (well, kinda). I made it in `<fc>` because it's just after the background color, it made sense to me but I'll move it to another tag.

jaor:
On Mon, Jul 27 2020, Unoqwy wrote:

> I'm a bit busy today, I'll fix the hints and update the readme tomorrow.
> The offset parameters on <fc> are meant for the background.

oh, i see. i misunderstood, i thought they were for the text. in that
case, if you feel it makes more sense to keep them close to the color in
fc than having a separate tag, or the latter complicates the
implementation considerably, please feel free to keep it as it is now.

Unoqwy:
Everything should be fine now. I think the readme is explicit enough, what I wrote about background offsets may be a bit confusing but all it takes is a try to understand what it does anyway.

jaor:
this looks good, thanks.  could we please rebase on top of current master?

jaor:
i've played with this a bit, btw.  for me, boxes don't draw fully in
view, but i know from previous experience that xft makes almost
impossible to get the font metrics right, and depends a lot on what
specific font one uses, and i just gave up trying to "fix" it. in my
case i think what's interferring is that is use different textOffsets,
so it's most probably a matter of me getting them right.

also, the default foreground for a box seems to be the background
color (?), which was surprising (i was expecting the foreground).

finally, it's honestly not clear to me what all the box types are
exactly, or how the alignment parameter interacts with them. i could
experiment more, but it's not very practical to try and see with so
many options.  so expect more people asking for more documentation :)

Unoqwy:
I fixed an issue with line width and rebased the branch. However, I wasn't able to reproduce your issue:


As you can see it works fine even with textOffsets.

The default box color is white, I can make it to be the foreground color if needed.
`BBTop`, `BBBottom`, `BBLeft` and `BBRight` are single borders while `BBVBoth` (meaning Box Border Vertical Both) groups top and bottom and `BBHBoth` groups left and right. BBFull groups all 4 borders.
The alignment parameter acts normally for top and bottom: alignment only matters when an offset is specified in which case `L` will have its offset on the right (because it's aligned to left, so the offset is on the opposite side), `C` will have offsets on both sides and `R` on the left. When it comes to left and right, `L` can be seen as `T` (top) and `R` as `B` (bottom). I'm not sure to be able to explain this well in the readme, maybe someone else could update it afterward?

jaor:
On Sat, Aug 01 2020, Unoqwy wrote:

> I fixed an issue with line width and rebased the branch. However, I
> wasn't able to reproduce your issue: As you can see it works fine even
> with textOffsets.

yes... i think what was happening in my case is that i use a border for
the full bar, and it's drawn on top of the box where they intersect.

> The default box color is white, I can make it to be the foreground
> color if needed.

i think it'd make more sense, specially if we want to say that we have
defaults that "do what's expected" (otherwise, people like me with white
backgrounds will just think  it doesn't work :))

> BBTop, BBBottom, BBLeft and BBRight are single borders while BBVBoth
> (meaning Box Border Vertical Both) groups top and bottom and BBHBoth
> groups left and right. BBFull groups all 4 borders.  The alignment
> parameter acts normally for top and bottom: alignment only matters
> when an offset is specified in which case L will have its offset on
> the right (because it's aligned to left, so the offset is on the
> opposite side), C will have offsets on both sides and R on the
> left. When it comes to left and right, L can be seen as T (top) and R
> as B (bottom). I'm not sure to be able to explain this well in the
> readme, maybe someone else could update it afterward?

i can probably write something, no problem. say i want a full box that
leaves a margin of 2 above and below the border of the full bar: i was
trying `<box=BBFull T 2 1 black>`, but all i see if i do that is an
empty xmobar (none of the monitors is shown, not even the ones outside
the box)... is that working for you?

Unoqwy:
>  say i want a full box that leaves a margin of 2 above and below the border of the full bar

Offsets are not margins, they depend on the alignment (for example `<box=BBBottom C 5 0 black>` will underline with a line smaller than the text length). Borders always draw with 0 margin (should I add a parameter for margins too?).

> i was trying `<box=BBFull T 2 1 black>`, but all i see if i do that is an empty xmobar 

`T` is not a valid alignment, only `L`, `C` and `R` are. I'm bad at explaining things but when I wrote about `T` I meant it "as if".

> i think it'd make more sense, specially if we want to say that we have defaults that "do what's expected" (otherwise, people like me with white backgrounds will just think it doesn't work :))

I didn't think about it, you're right.

If you want to see a working example, you can take a look at my dot files (I'm using boxes in 2 situations):
- [underline](https://github.com/Unoqwy/dotfiles/blob/207e106df0c9a80937d28fbf1263b537ff8de152/.xmonad/xmobar/xmobar.hs#L17)
- [top and bottom boxes with different colors and a full background](https://github.com/Unoqwy/dotfiles/blob/207e106df0c9a80937d28fbf1263b537ff8de152/.xmonad/xmonad.hs#L183)
(screenshots in the first post of this PR, slightly modified since then)

jaor:
On Sat, Aug 01 2020, Unoqwy wrote:

>     say i want a full box that leaves a margin of 2 above and below the border of the full bar
>    
> Offsets are not margins, they depend on the alignment (for example
> <box=BBBottom C 5 0 black> will underline with a line smaller than the
> text length).

width 0? if that's not a typo and causes a visible line to be drawn, i'm
very confused...

> Borders always draw with 0 margin (should I add a parameter for
> margins too?).

i think it'd be nice, although on the flip side the expression for the
box will gain yet another parameter, and it's already very hard to
understand (by now, i must confess i'm not sure i'll really be able of
documenting it properly).

Unoqwy:
> width 0? if that's not a typo and causes a visible line to be drawn
That's indeed a typo. The width specified here is the width given to XSetLineAttributes.

> the box will gain yet another parameter, and it's already very hard to understand
I think we shouldn't limit to a specific amount of parameters but instead maybe try to find another way of passing them? For example, would `<box=position:Full align:C offset:10 margin:2>` be a good solution to make it easier to understand/document? Also `position` may sound confusing because `Full` is not a position, maybe renaming `position` to `border` could help?

jaor:
On Sun, Aug 02 2020, Unoqwy wrote:

>     That's indeed a typo. The width specified here is the width given
>     to XSetLineAttributes.

phew :)

>     I think we shouldn't limit to a specific amount of parameters but
>     instead maybe try to find another way of passing them? For
>     example, would <box=position:Full align:C offset:10 margin:2> be a
>     good solution to make it easier to understand/document? Also
>     position may sound confusing because Full is not a position, maybe
>     renaming position to border could help?

yes, that's a good idea. perhaps we could use `type`, and something
like (the = alone by itself together with the :'s looks a bit weird to
me):

            <box type=Full align=C offset=1 margin=3 color=black>...</box>

or if you like colons:

            <box type:Full color:black align:C offset:1 margin:3>...</box>

with all the parameters optional and, ideally, given in any
order... what do you think?

Unoqwy:
> \<box type=Full align=C offset=1 margin=3 color=black>...\</box>
> with all the parameters optional and, ideally, given in any order... what do you think?

Sounds great, it will be easier to understand and use. I'm wondering if it wouldn't be better to remove the `align` parameter and move it in `offset` (`<box type=Bottom offset=5>` with `C` by default / `<box type=Bottom offset=10,L`) since it cannot be used without an offset anyway, what do you think?

I'll make the changes this week, probably Thursday or Friday.

jaor:
On Sun, Aug 02 2020, Unoqwy wrote:

> I'm wondering if it wouldn't be better to remove the align parameter
> and move it in offset (<box type=Bottom offset=5> with C by default /
> <box type=Bottom offset =10,L) since it cannot be used without an
> offset anyway, what do you think?

yes, since they have to always be specified together and a default
offset of 0 has no effect, that makes sense.  it could perhaps even be
`10L` etc., sans comma, although i don't have a strong opinion on that.

Unoqwy:
Done! :)

Changes:
- The color defaults to the foreground color
- All arguments are named, optional and in no specific order
- Added margins (`mt`, `mr`, `mb`, `ml` -> Top, Right, Bottom, Left)

Examples:
- `<box>foo</box>`
- `<box type=Bottom width=4>foo</box>`
- `<box type=Top color=blue offset=C10>foo</box>` (the alignment letter is placed **before**)
- `<box color=red mb=5 ml=2>foo</box>`

Unoqwy:
I feel like I messed up something with the rebase but I'm wondering what

Unoqwy:
Fixed it, next time I'll check twice before force-pushing

jaor:
On Fri, Aug 07 2020, Unoqwy wrote:

> I feel like I messed up something with the rebase but I'm wondering what

it's not rebased against current master, maybe you didn't have it
updated in your local repo?

jaor:
yes, now it merges cleanly.  but i get this warning when compiling:

```
src/Xmobar/X11/Draw.hs:202:3: warning: [-Wincomplete-patterns]
    Pattern match(es) are non-exhaustive
    In a case alternative:
        Patterns not matched:
            BBVBoth
            BBHBoth
            BBFull
    |
202 |   case pos of
```

Unoqwy:
Should be fixed.

jaor:
@Unoqwy i am trying to document this, but it's proving very difficult.  i confess i have no idea how `mt`, `mb`, etc. work, i never get the results i expect, and their meaning in the code is not evident for me either.  if you don't think you'd be able to document how they work either, how about we simply remove them? i worry about their maintainability.  what do you think?

Unoqwy:
They are all possible margins (`mt` stands for `margin-top`, `mb` stands for `margin-bottom`, `ml` stands for `margin-left` and `mr` stands for `margin-right`). I split them for easier understanding (instead of having something like `margin=0,0,2,0` that wouldn't be easy to read).
If you take `<box type=Bottom color=red mb=3>foo bar</box>` as an example, `foo bar` will be underlined in red but instead of being completely at the bottom of the bar, the red line will be 3 pixels above the absolute bottom, closer to the text.

jaor:
On Sun, Aug 09 2020, Unoqwy wrote:

> They are all possible margins (mt stands for margin-top, mb stands for margin-bottom, ml stands for
> margin-left and mr stands for margin-right). I split them for easier understanding (instead of having
> something like margin=0,0,2,0 that wouldn't be easy to read).

> If you take <box type=Bottom color=red mb=3>foo bar</box> as an
> example, foo bar will be underlined in red but instead of being
> completely at the bottom of the bar, the red line will be 3 pixels
> above the absolute bottom, closer to the text.

aaah, so these ones don't work with L,C,R prefixes, right? also, does it
make sense to have, say, type Top and mb? or type Left and mt? or is
it only one (or two) kind of alignment allowed per box type?

Unoqwy:
> aaah, so these ones don't work with L,C,R prefixes, right?

Yeah, L,C,R are only for the offset parameter, not margins.

> also, does it make sense to have, say, type Top and mb? or type Left and mt?

It does since a box can have up to 4 borders (Top, Right, Bottom and Left) at the same time and margins are individual.
Say you want a `VBoth` box with an even margin on top and bottom you can use `<box type=VBoth mt=3 mb=3 color=green>`. Furthermore, you can combine margins with offset: if we now want to add side borders to our green box that match the margins, we can nest another box like this `<box type=VBoth mt=3 mb=3 color=green><box type=HBoth offset=C3 color=green>Some text with a rectangular box around, not completely on top nor bottom</box></box>`

jaor:
okay, it's making sense now, thanks.  i've updated the readme.

Unoqwy:
I'm confused with a statement in the readme:
> For instance, `l=R5` asks for a
    margin of 5 pixels to the left, with right alignment.

Alignment cannot be use with margins, only with the offset as I said before (how would an alignment work with a margin?).
Also, you typed `l=` instead of `ml=`.

> For example, a box underlining its text with a red line of with 2

And here, `with` instead of `width`.
Other than that it's clear and complete :+1:

jaor:
On Sun, Aug 09 2020, Unoqwy wrote:

> I'm confused with a statement in the readme:
>
>     For instance, l=R5 asks for a
>     margin of 5 pixels to the left, with right alignment.

sorry, that was just copy/paste leftovers. removed.

Unoqwy:
I think you forgot to push the changes since the readme hasn't changed and is still misleading

jaor:
yes, was in the middle of pushing something else.  should be done now, thanks.

Issue #454, opened 2020-06-04
Surprising use of defaults and catch in `runVolumeWith`


wygulmage:
There are a couple of spots in `XMobar.Plugins.Monitors.Volume.runVolumeWith` where error handling seems to be done inconsistently. This may make be the correct behavior, but I can't figure out why. If it is correct, it would be good to add some documentation.

It centers on the helper function `channel'`:
```haskell
channel' :: PerChannel a -> a -> IO (Maybe a)
channel' v r = AE.catch 
    (getChannel FrontLeft v) 
    (const $ return $ Just r)
```
When the channel errors out, it returns `Just` the default value `r`, but when the channel does not exist, it returns `Nothing`.

`channel'` is used only in the helpers `getDB`, `getVal`, and `getSw`, which are in turn used only in the helper `readMixer`.
`readMixer` handles errors by returning `(Nothing, Nothing, Nothing, Nothing, Nothing)`.

```haskell
readMixer = AE.catch 
    (withMixer mixerName $ \mixer -> do
               control <- getControlByName mixer controlName
               (lo, hi) <- liftMaybe $ getRange <$> volumeControl control
               val <- getVal $ volumeControl control
               db <- getDB $ volumeControl control
               sw <- getSw $ switchControl control
               return (fmap toInteger lo, fmap toInteger hi, val, db, sw))
    (const $ return (Nothing, Nothing, Nothing, Nothing, Nothing))
```

The defaults of `getDB`, `getVal`, and `getSw` are only used when one of those functions throws an IO error after `getControlByName mixer controlName` and `liftMaybe $ getRange <$> volumeControl` succeed. Why is the default used in that rare case but not when the channel does not exist?

Three alternative behaviors for `channel'` make sense to me:
1. Return `Nothing` on an error, just like `readMixer` (for this you just get rid of the defaults):
```haskell
channel' :: PerChannel a -> IO (Maybe a)
channel' v = AE.catch 
    (getChannel FrontLeft v) 
    (const (return Nothing))
```
2. Use the default to unwrap the `Maybe`:
```haskell
channel' :: PerChannel a -> a -> IO a
channel' v r = AE.catch
    (fmap (fromMaybe r) (getChannel FrontLeft v))
    (const (return (Just r)))
```
3. Use the default value when the channel doesn't exist and return Nothing on an error (reverse of current behavior):
```haskell
channel' :: PerChannel a -> a -> IO (Maybe a)
channel' v r = AE.catch
[...]

PR #481, opened 2020-08-07
removed default arg of channel', channel (Plugins.Monitors.Volume)

wygulmage:
Per issue #454

Both functions had a default parameter for use in some error cases. Now each accepts only one parameter (a
PerChannel), and return Nothing on an error.

The definition of `channel` confused me, so I simplified it. Hopefully it's now
more clear that it just applies `toInteger` to the `IO (Maybe
CLong)` that `channel'` returns.


jaor:
thanks!

Issue #476, opened 2020-07-24
The fonts are not being rendered properly


shhra:
![image](https://user-images.githubusercontent.com/29188967/88383660-a85d5c00-cdca-11ea-805b-10e14cd12909.png)
As you can see the font awesome isn't being rendered properly, while on the another side it works perfectly fine.
My config to xmonad is as follows:

```
xmobarEscape = concatMap doubleLts
  where
        doubleLts '<' = "<<"
        doubleLts x   = [x]

myWorkspaces = clickable . map xmobarEscape $ ["\xf0a0", "\xf134", "\xf135", "\xf136", "\xf137", "\xf138", "\xf139", "\xf123", "\xf153"]
  where                                                                      
        clickable l = [ "<action=xdotool key super+" ++ show n ++ ">" ++ ws ++ "</action>" |
                      (i,ws) <- zip [1..9] l,                                        
                      let n = i ] 

```
The log hook is as follows:
```
        , logHook            = dynamicLogWithPP $ xmobarPP
                               {
                               -- primary bar
                               ppOutput            = \x -> hPutStrLn xmproc0 x  >> hPutStrLn xmproc1 x
                               , ppCurrent         = xmobarColor "#4c062b" "" . wrap "[-" "-]"  -- Current workspace in xmobar
                               , ppVisible         = xmobarColor "#4c062b" "" . wrap "-{" "}-"  -- Visible but not current workspace
                               , ppHidden          = xmobarColor "#851E1E" "" . wrap "*" "*"   -- Hidden workspaces in xmobar
                               , ppHiddenNoWindows = xmobarColor "#215c1f" ""                  -- Hidden workspaces (no windows)
                               , ppTitle           = xmobarColor "#104293" "" . shorten 10    -- Title of active window in xmobar
                               --, ppSep             = "<fc=#666666> | </fc>"                  -- Separators in xmobar
                               , ppUrgent          = xmobarColor "#C45500" "" . wrap "!" "!"  -- Urgent workspace
                               , ppOrder           = \(ws:l:t:ex) -> [l, ws]
                               }
```
And the font config at xmboar is as follows
```
-- http://projects.haskell.org/xmobar/
-- install xmobar with these flags: --flags="with_alsa" --flags="with_mpd" --flags="with_xft"  OR --flags="all_extensions"
-- you can find weather location codes here: http://weather.noaa.gov/index.html

Config { 
    font = "xft:SF Pro Display:size=12:antialias=true,FontAwesome:size=13"
[...]

PR #482, opened 2020-08-08
Do conditional utf8 encoding

psibi:
With this change, xmobar should respect the data which it gets. Xmobar
should just render the data instead of trying to encode it.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,8 @@ _Bug fixes_
 
   - Documentation fixes (Tomáš Janoušek)
   - Don't get confused by empty configuration dirs (fixes #412)
+  - Xft rendering: Avoid encoding to UTF8 on all scenarios. This
+    causes issue to StdinReader monitor when the handle wasn't binary.
 
 ## Version 0.35.1 (June, 2020)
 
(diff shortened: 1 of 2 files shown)


psibi:
@jaor This PR is ready for review and hopefully a merge! This has been tested by @slotThe and @shhra and confirmed that it's working for them. This fixes https://github.com/jaor/xmobar/issues/476

jaor:
excellent. thanks!

liskin:
Folks, this seems to be an ugly workaround for a double-utf8 encoding bug in xmonad-contrib that was introduced in https://github.com/xmonad/xmonad-contrib/pull/346. See my explanation in https://github.com/xmonad/xmonad-contrib/issues/338#issuecomment-779940855 and https://github.com/xmonad/xmonad-contrib/issues/338#issuecomment-779961072. We should fix this in xmonad -contrib and revert it here, ideally, as it's only needed with unreleased versions of xmonad-contrib. Thoughts?

jaor:
+1

liskin:
FYI: I just submitted a revert PR at https://github.com/jaor/xmobar/pull/539

PR #489, opened 2020-10-09
update libmpd to 0.9.2.0

o-schneider:
The update was needed to display MPD status with MPD 0.22+.


jaor:
not sure... force-pushing to the PR's branch has always worked for me.

PR #493, opened 2020-10-25
Fix for broken DateZone monitor

dino-:
There's a longstanding bug in timezone-olson that causes it to fail to
read some zoneinfo files (but not all, oddly). This was resolved in
timezone-olson-0.2.0 which can be built against by using a later
Stackage snapshot than master currently points to.

This fix pushes the snapshot up to lts-16.0 and also modifies the
cabal version range for timezone-olson to set 0.2.0 as the minimum.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,5 +1,5 @@
-# ghc 8.6.5
-resolver: lts-14.20
+# ghc 8.8.3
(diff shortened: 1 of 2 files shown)


jaor (approved):

jaor:
merged, thanks!

Issue #494, opened 2020-10-26
DST change not affecting Date plugin


loop54-christoffer:
My config includes a command

```haskell
Run Date "D: %B %d %A W%V" "date" 10
```

and while my system time has automatically reflected the local DST switch the night before last, my xmobar keeps ticking on with the pre-switch time. Am I doing something wrong or is this an oversight in the implementation?

jaor:
On Mon, Oct 26 2020, Christoffer wrote:

> My config includes a command
>
> Run Date "D: %B %d %A W%V" "date" 10
>
> and while my system time has automatically reflected the local DST
> switch the night before last, my xmobar keeps ticking on with the
> pre-switch time. Am I doing something wrong or is this an oversight in
> the implementation?

I think that the DST info is read only once on startup, so it's not
refreshed unless one restarts xmobar.  If that's really the case,
reading it over an over with every refresh would be perhaps a bit
overkill for an event that happens twice a year.  But i would have to
look closer at the code to confirm that's really what's going on.

slotThe:
On Mon, Oct 26 2020 00:41, jao wrote:
> I think that the DST info is read only once on startup, so it's not
> refreshed unless one restarts xmobar.

This seems to be what's going on.  The startup command for `Date` is
simply

    start (Date f _ r) cb = do
      t <- getCurrentTime
      zone <- getTimeZone t
      go zone
     where
      go zone = doEveryTenthSeconds r $ date zone f >>= cb

so `date` gets the time-zone (which includes things like DST) that was
current when the plugin was started.

loop54-christoffer:
Yeah, that makes sense.

Of course, the details of the affair is complicated, but I think most places that observe DST switch during the first few hours after midnight, so refreshing the time zone nightly should be an improvement over the current situation, though of course still not ideal.

Even the places where the DST switch does not happen just after midnight, the switch is likely to happen during a full hour, so updating the timezone at a minute past the hour would cover almost all remaining cases.

That said, the simplest thing is probably to get the timezone with every refresh, as overkill as it may seem.

0:
The current DST transition might be the first time most people encounter #456. Restarting xmobar was a reasonable workaround once I realized that it was showing the wrong time, but it's probably undesirable to have a clock that can't be trusted to always be correct.

jaor:
On Sat, Oct 31 2020, Dmitri Iouchtchenko wrote:

> The current DST transition might be the first time most people
[...]

PR #498, opened 2020-11-02
Revert "Optimize date plugin"

ivanbrennan:
Revert #456 because it prevents time zone changes from being picked up without a restart.

That PR changed the date plugin to avoiding calling `getTimeZone` on each execution, instead calling it just once upon startup and reusing the zone for each subsequent execution. As a result, time zone changes (e.g. daylight savings shifts) are not detected without restarting xmobar.

I noticed this after my local time zone had shifted from EDT to EST. My xmobar showed 4:30 when the local time was in fact 3:30 (and running date on the command line confirmed that my system clock was actually aware of this shift). I had to restart xmobar in order to pick up the new time zone.

I repro'd the unexpected behavior by temporarily disabling my system's time syncing, setting the time to 30 seconds before the zone shift, running date to confirm I'd set the correct time, restarting xmobar, and observing.
```sh
sudo systemctl stop systemd-timesyncd.service
date --set="01:59:30"
date
```
I observed my xmobar clock go from 1:59 to 2:00, rather than from 1:59 to 1:00 as expected.

Following the same steps, I was able to verify that reverting #456 fixes the issue.

### DIFF
--- a/src/Xmobar/Plugins/Date.hs
+++ b/src/Xmobar/Plugins/Date.hs
@@ -31,16 +31,8 @@ data Date = Date String String Int
 
 instance Exec Date where
     alias (Date _ a _) = a
+    run   (Date f _ _) = date f
     rate  (Date _ _ r) = r
-    start (Date f _ r) cb = do
-      t <- getCurrentTime
-      zone <- getTimeZone t
-      go zone
-     where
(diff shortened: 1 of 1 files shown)


jaor:
Thanks @ivanbrennan for the detailed and careful report. @psibi this was the easy half of your proposal in #494 :)

PR #500, opened 2020-11-04
Fix #499 - Create a SetAlpha Signal and expose it over DBus

### LINKED ISSUE #499: Conditional and/or auto-updating alpha?
So I'm new to xmobar and I was wondering if there is a way to update the alpha of xmobar at runtime? I know we can specify the `alpha` parameter in the config and that works when then restarting xmobar and reloading the config but is there a way to change the alpha without restarting xmobar? like triggering a redraw somehow?

Additionally, assuming the above is possible, is there a way to change the alpha based on a condition and then triggering the alpha update? I currently have some workspace widgets being "injected" into xmobar using Run UnsafeStdInReader and part of that actually includes the current selected xmonad layout. I'm thinking it would be cool to parse the text coming to identify the current layout, and then update the xmobar bg alpha accordingly e.g. have it be transparent for some layouts and have it show the bgcolour if a full-screen layout is chosen.

Is this possible in any way?

Sorry for the wall of text, great project by the way!

linuxhype:
This is subject to the DBus limitation of only one xmobar process "holding the dbus" but with this PR it is possible to send a `SetAlpha x` where x is the Integer value of the desired Alpha (ranging between 0-255 as defined in xmobar config).

### DIFF
--- a/src/Xmobar/App/EventLoop.hs
+++ b/src/Xmobar/App/EventLoop.hs
@@ -176,6 +176,7 @@ eventLoop tv xc@(XConf d r w fs vos is cfg) as signal = do
 
          Hide   t -> hide   (t*100*1000)
          Reveal t -> reveal (t*100*1000)
+         SetAlpha t -> eventLoop tv xc { config = cfg { alpha = t }} as signal
          Toggle t -> toggle t
 
          TogglePersistent -> eventLoop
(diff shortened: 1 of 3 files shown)


jaor:
This goods nice and simple, thanks! Would you mind adding a brief mention of the new capability to the dbus section of the readme? (or i can do it tomorrow if you don't have the bandwidth).

jaor:
i've pushed in 082ef98 a variant of this PR, with a bit of
documentation.  thanks a lot for your contribution and patience!

PR #501, opened 2020-11-07
Changelog entry for Revert "Optimize date plugin"

liskin:
A missing bit from https://github.com/jaor/xmobar/pull/498 by @ivanbrennan.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,11 @@ _New features_
 
   - New command line option `--add-font` (Ivan Brennan)
 
+_Bug fixes_
+
+  - Fix date plugin not picking up DST and timezone changes
+    (Fixed by reverting the optimization merged in 0.34).
+
 ## Version 0.36 (August, 2020)
 
 _New features_


changelog.md
```diff
@@ -40,9 +40,6 @@ _New features_
   - New plugin `HandleReader` for reading data from a Haskell `Handle`. This is
     useful if you are running xmobar from within a Haskell program.
   - Build with ghc 8.10 allowed.
-  - Optimize date plugin by avoiding calling getTimeZone for each of
-    the time the date has to be updated. Instead, it's computed once
-    at the start and re-used for each invocation.
```
ivanbrennan, line 45: I wasn't sure whether it made sense to remove this, since it made it into a release (0.34). Would it be better to add a changelog entry noting the reversion?

ivanbrennan: If I recall correctly, I had intended to add a changelog entry noting the reversion. But then I forgot to do so :P

liskin: Oh, right, I probably wasn't thinking straight this morning. What you're saying makes more sense.

ivanbrennan (approved):
Thanks!

jaor:
thanks guys, merged (even though i bet @psibi will overwrite it when he comes with a better fix :))

PR #502, opened 2020-11-09
Add DateWithTimeZone constructor for computing Date with fixed time zones

psibi:
Also, this time I have added code for benchmarking both of the
components and this is the result in my machine:

```
benchmarked Date Benchmarks/Date
time                 4.434 μs   (4.397 μs .. 4.460 μs)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 4.439 μs   (4.413 μs .. 4.525 μs)
std dev              134.5 ns   (35.90 ns .. 281.0 ns)
variance introduced by outliers: 13% (moderately inflated)

benchmarked Date Benchmarks/DateWithTimeZone
time                 2.266 μs   (2.260 μs .. 2.271 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 2.258 μs   (2.254 μs .. 2.261 μs)
std dev              13.25 ns   (11.39 ns .. 15.34 ns)
```

And that indicates that `DateWithTimeZone` is around 48% faster than the `Date` monitor.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1548,11 +1548,11 @@ will display "N/A" if for some reason the `date` invocation fails.
 - Accepts same input as the above `Date` monitor.
 - Sample usage: `Run DateWithTimeZone "%a %b %_d %Y <fc=#ee9a00>%H:%M:%S</fc>" "date" 10`
 - The difference with the `Date` monitor is that it's better optimized
-  and 45% faster than the it. It computes the initial time zone
-  information at the start of the xmobar program and reuses it for the
-  subsequent invocations. That means this monitor isn't suited if you
-  are travelling among different time zones, have DST etc unless you
-  are ready to restart xmobar.
+  and 45% faster. It computes the initial time zone information at the
+  start of the xmobar program and reuses it for the subsequent
+  invocations. That means this monitor isn't suited if you are
+  travelling among different time zones, have DST etc unless you are
+  ready to restart xmobar.
 
 ## `DateZone Format Locale Zone Alias RefreshRate`
 
--- a/readme.md
+++ b/readme.md
@@ -65,6 +65,7 @@
     - [`StdinReader`](#stdinreader)
     - [`UnsafeStdinReader`](#unsafestdinreader)
     - [`Date Format Alias RefreshRate`](#date-format-alias-refreshrate)
+    - [`DateWithTimeZone Format Alias RefreshRate`](#datewithtimezone-format-alias-refreshrate)
     - [`DateZone Format Locale Zone Alias RefreshRate`](#datezone-format-locale-zone-alias-refreshrate)
     - [`CommandReader "/path/to/program" Alias`](#commandreader-pathtoprogram-alias)
     - [`PipeReader "default text:/path/to/pipe" Alias`](#pipereader-default-textpathtopipe-alias)
@@ -1541,6 +1542,17 @@ will display "N/A" if for some reason the `date` invocation fails.
 - Format is a time format string, as accepted by the standard ISO C
   `strftime` function (or Haskell's `formatCalendarTime`).
 - Sample usage: `Run Date "%a %b %_d %Y <fc=#ee9a00>%H:%M:%S</fc>" "date" 10`
+
+## `DateWithTimeZone Format Alias RefreshRate`
+
+- Accepts same input as the above `Date` monitor.
+- Sample usage: `Run DateWithTimeZone "%a %b %_d %Y <fc=#ee9a00>%H:%M:%S</fc>" "date" 10`
+- The difference with the `Date` monitor is that it's better optimized
+  and 45% faster than the it. It computes the initial time zone
+  information at the start of the xmobar program and reuses it for the
+  subsequent invocations. That means this monitor isn't suited if you
+  are travelling among different time zones, have DST etc unless you
+  are ready to restart xmobar.
 
 ## `DateZone Format Locale Zone Alias RefreshRate`
 
--- a/src/Xmobar/Plugins/Date.hs
+++ b/src/Xmobar/Plugins/Date.hs
@@ -17,7 +17,7 @@
 --
 -----------------------------------------------------------------------------
 
-module Xmobar.Plugins.Date (Date(..)) where
+module Xmobar.Plugins.Date (Date(..), date, dateWithTimeZone) where
 
 import Xmobar.Run.Exec
 
@@ -27,12 +27,32 @@ import System.Locale
 import Data.Time
 
 data Date = Date String String Int
+          | DateWithTimeZone String String Int
     deriving (Read, Show)
 
 instance Exec Date where
     alias (Date _ a _) = a
+    alias (DateWithTimeZone _ a _) = a
     run   (Date f _ _) = date f
+    run   (DateWithTimeZone f _ _) = do
+                     t <- getCurrentTime
+                     zone <- getTimeZone t
+                     dateWithTimeZone zone f
     rate  (Date _ _ r) = r
+    rate  (DateWithTimeZone _ _ r) = r
+    start (Date f _ r) cb = doEveryTenthSeconds r $ (date f) >>= cb
+    start (DateWithTimeZone f _ r) cb = do
+      t <- getCurrentTime
+      zone <- getTimeZone t
+      go zone
+     where
+      go zone = doEveryTenthSeconds r $ dateWithTimeZone zone f >>= cb
 
 date :: String -> IO String
 date format = fmap (formatTime defaultTimeLocale format) getZonedTime
+
+dateWithTimeZone :: TimeZone -> String -> IO String
+dateWithTimeZone timezone format = do
+  time <- getCurrentTime
+  let zonedTime = utcToZonedTime timezone time
+  pure $ formatTime defaultTimeLocale format zonedTime
--- a/src/Xmobar/Plugins/Date.hs
+++ b/src/Xmobar/Plugins/Date.hs
@@ -40,7 +40,7 @@ instance Exec Date where
                      dateWithTimeZone zone f
     rate  (Date _ _ r) = r
     rate  (DateWithTimeZone _ _ r) = r
-    start (Date f _ r) cb = doEveryTenthSeconds r $ (date f) >>= cb
+    start (Date f _ r) cb = doEveryTenthSeconds r $ date f >>= cb
     start (DateWithTimeZone f _ r) cb = do
       t <- getCurrentTime
       zone <- getTimeZone t
--- a/src/Xmobar/Plugins/Date.hs
+++ b/src/Xmobar/Plugins/Date.hs
@@ -33,11 +33,6 @@ data Date = Date String String Int
 instance Exec Date where
     alias (Date _ a _) = a
     alias (DateWithTimeZone _ a _) = a
-    run   (Date f _ _) = date f
-    run   (DateWithTimeZone f _ _) = do
-                     t <- getCurrentTime
-                     zone <- getTimeZone t
-                     dateWithTimeZone zone f
     rate  (Date _ _ r) = r
     rate  (DateWithTimeZone _ _ r) = r
     start (Date f _ r) cb = doEveryTenthSeconds r $ date f >>= cb
--- a/bench/main.hs
+++ b/bench/main.hs
@@ -1,26 +1,51 @@
-{-#LANGUAGE RecordWildCards#-}
+{-# LANGUAGE RecordWildCards #-}
 
-import Gauge
-import Xmobar
-import Xmobar.Plugins.Monitors.Common.Types
-import Xmobar.Plugins.Monitors.Common.Run
-import Xmobar.Plugins.Monitors.Cpu
 import Control.Monad.Reader
 import Data.IORef (newIORef)
+import Data.Time
+import Gauge
+import Xmobar
+import Xmobar.Plugins.Monitors.Common.Run
+import Xmobar.Plugins.Monitors.Common.Types
+import Xmobar.Plugins.Monitors.Cpu
+
+data BenchmarkInput = BenchmarkInput
+  { cpuInput :: CpuArguments,
+    dateInput :: DateInput
+  }
+
+data DateInput = DateInput
+  { diFormat :: String,
+    diZone :: TimeZone
+  }
 
 main :: IO ()
 main = do
   cpuParams <- mkCpuArgs
-  defaultMain $ normalBench cpuParams
-    where
-      normalBench args = [ bgroup "Cpu Benchmarks" $ normalCpuBench args]
+  time <- getCurrentTime
+  zone <- getTimeZone time
+  let benchInput =
+        BenchmarkInput
+          { cpuInput = cpuParams,
+            dateInput =
+              DateInput
+                { diFormat = "D: %B %d %A W%V",
+                  diZone = zone
+                }
+          }
+  defaultMain $ normalBench benchInput
+  where
+    normalBench args =
+      [ -- bgroup "Cpu Benchmarks" $ normalCpuBench (cpuInput args)
+        bgroup "Date Benchmarks" $ dateBench (dateInput args)
+      ]
 
 runMonitor :: MConfig -> Monitor a -> IO a
 runMonitor config r = runReaderT r config
 
 mkCpuArgs :: IO CpuArguments
-mkCpuArgs = getArguments ["-L","3","-H","50","--normal","green","--high","red", "-t", "Cpu: <total>%"]
-  
+mkCpuArgs = getArguments ["-L", "3", "-H", "50", "--normal", "green", "--high", "red", "-t", "Cpu: <total>%"]
+
 -- | The action which will be benchmarked
 cpuAction :: CpuArguments -> IO String
 cpuAction = runCpu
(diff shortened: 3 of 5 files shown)


readme.md
```diff
@@ -1542,6 +1543,17 @@ will display "N/A" if for some reason the `date` invocation fails.
   `strftime` function (or Haskell's `formatCalendarTime`).
 - Sample usage: `Run Date "%a %b %_d %Y <fc=#ee9a00>%H:%M:%S</fc>" "date" 10`
 
+## `DateWithTimeZone Format Alias RefreshRate`
+
+- Accepts same input as the above `Date` monitor.
+- Sample usage: `Run DateWithTimeZone "%a %b %_d %Y <fc=#ee9a00>%H:%M:%S</fc>" "date" 10`
+- The difference with the `Date` monitor is that it's better optimized
+  and 45% faster than the it. It computes the initial time zone
```
jaor, line 1551: typo: "than the it" can be just deleted

jaor (changes requested):
This looks excellent.  It needs a rebase, though, and you can take the chance to fix a typo i saw bellow.  Thanks a lot!

src/Xmobar/Plugins/Date.hs
```diff
@@ -27,12 +27,32 @@ import System.Locale
...
     deriving (Read, Show)
 
 instance Exec Date where
     alias (Date _ a _) = a
+    alias (DateWithTimeZone _ a _) = a
     run   (Date f _ _) = date f
+    run   (DateWithTimeZone f _ _) = do
+                     t <- getCurrentTime
+                     zone <- getTimeZone t
+                     dateWithTimeZone zone f
```
liskin, line 40: This isn't needed I believe. If `start` is implemented, `run` is never used, so this code would just confuse people.

liskin:
Yeah, the code is nice and all, but the numbers don't add up. There's simply no way shaving off 2 μs off something happening once a second is going to result in an improvement of several percent of xmobar CPU usage:

```
$ units --terse '4.434 μs / 1 second' percent
0.0004434
$ units --terse '2.266 μs / 1 second' percent
0.0002266
```

And this is already in **percents**. As a fraction it's two more zeros in front of it. Remember, a μs is a thousandth of a millisecond. When top shows xmobar is taking 1 percent of CPU time, that's 10 milliseconds. 2 μs improvement is 1/10000 of that. I'm sorry to say it this way, but this really feels like the proverbial premature optimization that is said to be the root of all evil. :-)

@psibi can you please re-check that you're seeing measurable differences between an xmobar running the Date plugin at once per second refresh (that would be `rate = 10`, not `rate = 1`) and an xmobar running the DateWithTimeZone plugin at once per second refresh?

psibi:
@liskin My point in introducing the whole `DateWithTimeZone` is to
avoid an additional FFI call overhead which is happening:
https://www.stackage.org/haddock/lts-16.20/time-1.9.3/src/Data-Time-LocalTime-Internal-TimeZone.html#getTimeZone

While what you are saying is fine and I agree, my point is that it
does introduce perf improvments. Note that I'm still not satisfied
with my xmobar's performance and I'm planning to improve it
incrementally as time allows. (But as you have mentiond earlier, it
might be an issue with Xft rendering and I'm also planning to look
into that when I get some spare time)

And from observing htop's usage only with Date plugin - yes my perf
seems to have improved.

src/Xmobar/Plugins/Date.hs
psibi: I think it's the other way round. If `run` is implemented, then `start` is not needed as the default implementation is like this:

```
    run     :: e -> IO String
    run     _    = return ""
    start   :: e -> (String -> IO ()) -> IO ()
    start   e cb = go
        where go = doEveryTenthSeconds (rate e) $ run e >>= cb
```

I find it confusing, so I have implement all the functions. But I can remove the `start` method implementation - if you feel strongly.

liskin:
@psibi I'm not disputing that there is a non-zero overhead of getTimeZone. Of course there is *some* overhead. But we need to look at this overhead in the context of xmobar as a whole!

So what I'm trying to say is there's a tradeoff between performance and complexity of code. It's not worth making the code more complex if the improvement isn't measurable in the context of the whole thing. You can justify absolutely anything with a microbenchmark (pun intended :smile:) like this, but you need to step further out and look at the larger picture.

Can you post screenshots of htop with xmobar with Date and xmobar with DateWithTimeZone? I have a really hard time believing there's a difference, because from the numbers you posted here there simply can't be any measurable one. If there is one, we're missing something, and I'd love to understand what that missing thing is before merging something we don't really understand.

psibi:
@liskin Thanks for the feedback!

> So what I'm trying to say is there's a tradeoff between performance and complexity of code. 

I guess I don't find the code complex mostly because I have authored it.  :grinning: 

> Can you post screenshots of htop with xmobar with Date and xmobar with DateWithTimeZone?

Sure, but that is non scientific and I don't want to rely on that. Even though observing my xmobar via htop was what led to this MR. And I find the above `gauge` bencmarks slightly better than that.

Anyway, I totally understand if this MR isn't merged and would leave the decision to you and @jaor  as I could just rely on my local patches for this. :-)

readme.md
psibi: @jaor Rebased and fixed the typo. Thanks!

src/Xmobar/Plugins/Date.hs
liskin: No, if you only implement `run`, you'll end up where you started, with `getTimeZone` being called again and again. When you need something more complex, like an upfront initialization, you need to override `start`.

See? You already managed to confuse yourself. :-)
Now imagine someone who didn't write this code trying to understand it.
Please remove the `run` implementation.

psibi: @liskin Lol! Yeah, I manged to confuse myself. :-) Thanks for pointing it out, have pushed a fix removing it.

liskin:
> Sure, but that is non scientific and I don't want to rely on that.

You're absolutely right, so let's go figure out something more reliable then! :-)

How about this: `$ time timeout 1h xmobar`
This will run xmobar for an hour, then report the overall CPU time in userspace and system. Run that once for xmobar with Date and once for xmobar with DateWithTimeZone. Maybe not an hour, `10m` should be enough.

psibi:
> How about this: $ time timeout 1h xmobar

@liskin Yeah, that does sound more reasonable. Let me do that and get back to you. Infact, I'm planning to do with other extensions enabled in both the cases. Let me know if you think that's a bad idea. :-)

liskin:
> Infact, I'm planning to do with other extensions enabled in both the cases. Let me know if you think that's a bad idea. :-)

Yes, to look at the whole thing, enabling other extensions is a good idea.

But it'd also be interesting to see how it performs isolated, with xmobar only checking the time and rendering it every second. If my theory is correct, we shouldn't see a measurable difference in such a setup either.

psibi:
Cool. As a first step, let me check with other extensions enabled then as it's much easier for me to it. I will report back after couple of hours as I'm planning to run it for 1 hour each. :-)

liskin:
> I guess I don't find the code complex mostly because I have authored it.

Yeah, that's a factor. Also, it's not just code, you're adding additional options for users, more documentation for them to read, etc. If it turns out there is little measurable difference, we would potentially have users wasting time switching between Date and DateWithTimeZone and being confused why it's not making their xmobars consume less CPU time when in fact their problems are elsewhere. So that's primarily why I'm making your life harder here. :smiley:

I hope it doesn't look like useless nitpicking. It's just... the numbers so far weren't convincing and I'd really really like to understand what's going on here.

psibi:
@liskin These are the results from my end:

This is the output:

-- DateWithTimeZone

```
Command exited with non-zero status 124
9.06user 1.13system 1:00:00elapsed 0%CPU (0avgtext+0avgdata 34424maxresident)k
0inputs+0outputs (0major+6163minor)pagefaults 0swaps
```

-- Date

```
Command exited with non-zero status 124
9.67user 1.67system 1:00:00elapsed 0%CPU (0avgtext+0avgdata 34600maxresident)k
0inputs+0outputs (0major+6269minor)pagefaults 0swaps
```

liskin:
Interesting. If I'm interpreting the numbers right, that means xmobar with Date used (9.67 + 1.67) / 3600 = 0.00315, that is 3.15 milliseconds of CPU time every second, whereas xmobar with DateWithTimeZone used (9.06 + 1.13) / 3600 = 0.00283, that is 2.83 milliseconds every second. So that's a difference of 320 microseconds, or roughly 10 %.

That is less than the 45 % one might expect from reading the readme, but it's also significantly more than the 3 microseconds one would expect from the microbenchmark.

I can think of two explanations:

1. getTimeZone in a tight loop benefits from some form of caching somewhere (in glibc, in kernel, perhaps both) and is considerably more expensive when called once a second
2. this is a measurement error; after all this was your normal xmonad instance so if you used the computer differently during these two measurements, xmonad would refresh a different number of times (switching workspaces, windows, etc.) and the results would reflect your usage pattern more than the time spent getting timezones; additionally, CPU frequency scaling might be involved (whenever your computer is busy, CPU runs at higher frequency and thus xmonad consumes less time)

I'll try to run a similar measurement overnight to rule out number 2.

liskin:
I found a way to avoid measurement errors even without running it overnight: just run two xmobars at the same time. :-)

So far my results are:

```
[tomi@notes tmp]$ time timeout 1h xmobar .xmobarrc-date
xmobar: Caught signal 15; exiting...

real    60m0,005s
user    0m15,599s
sys     0m6,001s

[tomi@notes tmp]$ time timeout 1h xmobar .xmobarrc-datewithtimezone
xmobar: Caught signal 15; exiting...

real    60m0,008s
user    0m15,324s
sys     0m6,191s

; (15.599 + 6.001) / 3600
        0.006
; (15.324 + 6.191) / 3600
        ~0.00597638888888888889
; 0.006 - 0.00597638888888888889
        0.00002361111111111111
; (0.006 - 0.00597638888888888889) * 1000000
        23.61111111111111
```

That is, my xmobar is active 6ms out of every 1s, and switching to DateWithTimeZone makes that 5.976ms, which is a difference of 23.6μs. Still more than 2.2μs from the μbenchmark, but a bit more believable. I'll now strip my config to just a date and run it overnight to confirm this.

Would you be able to repeat your experiment with two xmobars at the same time?
I used this in xmobarrc:
```
, position = Static{xpos = 0, ypos = 15, width = 1920, height = 15}
```
and the other one at `ypos = 30`.

psibi:
> That is, my xmobar is active 6ms out of every 1s, and switching to DateWithTimeZone makes that 5.976ms, which is a difference of 23.6μs. Still more than 2.2μs from the μbenchmark, but a bit more believable. I'll now strip my config to just a date and run it overnight to confirm this.

Thanks for the report, but it's hard for me to interpret the result. Has there been any significant improvements or it has been opposite case from your observations ?

> Would you be able to repeat your experiment with two xmobars at the same time?

Sure, I could do that. But I have to modify my configuration to strip all the other monitors. I don't run my xmobar via the configuration file - so it would take some time. Either wise, I would be happy to see your result for the same and reach us to the conclusion!

liskin:
> Thanks for the report, but it's hard for me to interpret the result. Has there been any significant improvements or it has been opposite case from your observations ?

I'm going to do one additional measurement. The one with just date running overnight for 6 hours resulted in this:

```
    [tomi@notes tmp]$ time timeout 6h xmobar .xmobarrc-date
    xmobar: Caught signal 15; exiting...

    real    360m0,008s
    user    0m11,992s
    sys     0m5,257s

    [tomi@notes tmp]$ time timeout 6h xmobar .xmobarrc-datewithtimezone
    xmobar: Caught signal 15; exiting...

    real    360m0,007s
    user    0m11,023s
    sys     0m5,025s

    ; (11.992 + 5.257) / (360 * 60)
            ~0.00079856481481481481
    ; (11.023 + 5.025) / (360 * 60)
            ~0.00074296296296296296
    ; (0.00079856481481481481 - 0.00074296296296296296) * 1000000
            55.60185185185185
```

This means that a bare Date-only xmobar uses around 0.8 ms every second, and switching to DateWithTimeZone makes that 0.74 ms, which is cca 7 % improvement. The additional measurement I'm going to do now is to run xmobar with DateWithTimeZone and another xmobar with both DateWithTimeZone and Date, in an attempt to figure out how much time Date needs, so that we can confirm or adjust the `45 %` in the documentation.

Anyway, the 7 % number I got is similar to your 10 %, so I'm starting to think that this may indeed be worth merging after all. :-)

psibi:
Thanks for the update and I would be eager to see your next results. And I think we should also document this strategy of benchmarking as I feel it would be quite handy when trying to optimise a specific monitor in Xmobar.

liskin:
> And I think we should also document this strategy of benchmarking as I feel it would be quite handy when trying to optimise a specific monitor in Xmobar.

That is an excellent point indeed. :-)

liskin:
```
[tomi@notes tmp]$ time timeout 2h xmobar .xmobarrc-datewithtimezone+date
xmobar: Caught signal 15; exiting...

real    120m0,008s
user    0m4,440s
sys     0m1,725s

[tomi@notes tmp]$ time timeout 2h xmobar .xmobarrc-datewithtimezone
xmobar: Caught signal 15; exiting...

real    120m0,009s
user    0m3,663s
sys     0m1,686s

; (4.440 + 1.725) / (120 * 60)
		0.00085625
; (3.663 + 1.686) / (120 * 60)
		~0.00074291666666666667
; (0.00085625 - 0.00074291666666666667) * 1000000
		113.33333333333333
```

So this means adding an additional Date needs 113 μs.
If DateWithTimeZone is cca 55 μs faster, that means it is indeed about 45 % faster.

It is, however, still worth noting that xmobar with just a date widget still needs at least 740 μs to refresh and render (on my hardware), so switching from Date to DateWithTimeZone is never going to bring real 45 % improvement, but around 10 % at best.

But I got another idea when I went for a bike ride: why don't we take the best of both worlds — refresh timezone once a minute or something like that. If we do that precisely when the seconds hand is :00, DST will probably work correctly in most real world cases. This would take 58 * 59 + 113 * 1 = 3535 μs every minute (Date would take 113 * 60 = 6780, DateWithTimeZone would take 3480), so this compromise is only 1.5 % slower than DateWithTimeZone, and that seems acceptable.

What do you think?

(This was a lovely exercise in absolute perfectionism, wasn't it? :-))

liskin:
A first stab at implementing that: https://github.com/liskin/xmobar/commit/e1932f31dc02a2b49db20609f7528a40ea48be3c

I don't like the code being so long and complex. But at least it would be completely transparent to users. Will report measurement once I have those.

psibi:
@liskin 

> why don't we take the best of both worlds — refresh timezone once a minute or something like that.

I don't have strong opinion on this, but in spite of that I would like
to still have DateWithTimeZone. What do you think about that ? Mostly
because:

- I still don't want to pay for the extra overhead. (1.5% according to you)
- The timezone I'm in, I don't have any DST issues. Also my usage
  involves daily starting and stopping my machines.

> If we do that precisely when the seconds hand is :00 (which is just time - mod time 60), DST will probably work correctly in most real world cases.

Probably. I don't have much idea about DST to comment on it and I feel
at this stage the implementation is becoming more complex and that
kind of makes me want to lean towards no. Probably a new constructor
to implement this logic `DateWithRefreshedTimeZone` ? Looking at your implementation, it seems you have done similarly using a new constructor.

> (This was a lovely exercise in absolute perfectionism, wasn't it? :-))

Yeah! :-) 

> I don't like the code being so long and complex. But at least it would be completely transparent to users. 

Yeah, it's complex. But IMO it's fine!

liskin:
> it seems you have done similarly using a new constructor

That's only to make it easier to compare measurements. My intention is to get this to be as almost as fast as your DateWithTimeZone was and then just replace the original Date implementation with it, so that the change is completely transparent to users.

Anyway I need to make a few additional measurements. Will report in another day or so.

liskin:
Oh and I just found out how to simplify the implementation: just use a separate `doEveryTenthSeconds` thread.
https://github.com/liskin/xmobar/commit/0e874122ca2996e55aeda7a531e25a03493c3307

Now let's hope this is fast enough to not need any configuration! :-)

psibi:
@liskin It would be interesting to see what the micro benchmark says also. :-)

liskin:
The night's results are in:

    [tomi@notes tmp]$ time timeout 6h xmobar .xmobarrc-date
    real    360m0,010s
    user    0m9,867s
    sys     0m4,644s
    
    [tomi@notes tmp]$ time timeout 6h xmobar .xmobarrc-dateopt
    real    360m0,008s
    user    0m9,535s
    sys     0m4,327s
    
    [tomi@notes tmp]$ time timeout 6h xmobar .xmobarrc-dateopt-10m
    real    360m0,010s
    user    0m9,780s
    sys     0m4,215s
    
    [tomi@notes tmp]$ time timeout 6h xmobar .xmobarrc-datewithtimezone
    real    360m0,006s
    user    0m9,658s
    sys     0m4,166s

```
; (9.867 + 4.644) / (360 * 60)
        ~0.00067180555555555556
; (9.535 + 4.327) / (360 * 60)
        ~0.00064175925925925926
; (9.780 + 4.215) / (360 * 60)
        ~0.00064791666666666667
; (9.658 + 4.166) / (360 * 60)
        0.00064
```

Seems we're getting close to an immeasurable difference between DateOpt and DateWithTimeZone...
I'll try the μbechmark (which will only reflect the difference between TimeZone and IORef TimeZone) next.

liskin:
https://github.com/liskin/xmobar/commit/734163a4abb2f808ac64ae495d0534572abcb07c

```
[tomi@notes .xmonad]$ stack build --bench --benchmark-arguments '--small Date' xmobar
xmobar> benchmarks
Running 1 benchmarks...
Benchmark xmobarbench: RUNNING...
Date Benchmarks/Date                     mean 5.017 μs  ( +- 167.9 ns  )
Date Benchmarks/DateOpt                  mean 2.833 μs  ( +- 34.51 ns  )
Date Benchmarks/DateWithTimeZone         mean 2.834 μs  ( +- 19.59 ns  )
Benchmark xmobarbench: FINISH
```

I'll clean this up and submit as a PR then. :-)

psibi:
Awesome!

liskin:
See https://github.com/jaor/xmobar/pull/503

psibi:
Closing this as this has been superseded by https://github.com/jaor/xmobar/pull/503

PR #503, opened 2020-11-12
Optimize Date plugin again (refresh timezone only once a minute)

liskin:
This makes the Date plugin approximately twice as fast, and makes xmobar
up to about 5–10 % faster if Date is the only active plugin. (If more
expensive plugins like Network or MultiCpu are used, it doesn't make any
measurable difference.)

Micro-benchmark results on my HW:

    Date Benchmarks/Date                     mean 2.833 μs  ( +- 16.08 ns  )
    Date Benchmarks/DateZonedTime            mean 5.020 μs  ( +- 32.91 ns  )
    Date Benchmarks/DateWithTimeZone         mean 2.827 μs  ( +- 20.52 ns  )

(DateZonedTime is the original implementation and DateWithTimeZone is
the implementation we had since 0.34 which never refreshes timezone.)

Real-life measurements (done overnight on an idle laptop, with all
measured xmobars running in parallel to ensure comparable conditions;
xmobars configured to only display date and with rate 10 — once per
second):

    $ time timeout 6h xmobar .xmobarrc-DateZonedTime
    real    360m0,010s
    user    0m9,867s
    sys     0m4,644s

    (9.867 + 4.644) / (360 * 60) = 0.000672

    $ time timeout 6h xmobar .xmobarrc-Date
    real    360m0,008s
    user    0m9,535s
    sys     0m4,327s

    (9.535 + 4.327) / (360 * 60) = 0.000642

    $ time timeout 6h xmobar .xmobarrc-Date-10m
    real    360m0,010s
    user    0m9,780s
    sys     0m4,215s

    (9.780 + 4.215) / (360 * 60) = 0.000648

    $ time timeout 6h xmobar .xmobarrc-DateWithTimeZone
    real    360m0,006s
    user    0m9,658s
    sys     0m4,166s

    (9.658 + 4.166) / (360 * 60) = 0.000640

(.xmobarrc-Date-10m is the proposed implementation, but with timezone
refresh every 10 minutes instead of every 1 minute)

Interpretation of these results:

 * refreshing xmobar with just date takes around 650 μs
 * that is xmobar with just date uses around 0.065 % of CPU time
 * refreshing timezone takes additional cca 30 μs

When we only refresh timezone once a minute, these 30 μs become 0.5 μs
amortized, and that should be acceptable to even the most dedicated
perfectionist :-)

Fixes: a58e32f7c8af ("Revert "Optimize date plugin"")
Fixes: 878db3908060 ("Optimize date plugin")
Co-authored-by: @psibi

---

This replaces https://github.com/jaor/xmobar/pull/502.

### DIFF
--- a/bench/main.hs
+++ b/bench/main.hs
@@ -1,32 +1,19 @@
-{-#LANGUAGE RecordWildCards#-}
+module Main (main) where
 
 import Gauge
 import Xmobar
-import Xmobar.Plugins.Monitors.Common.Types
-import Xmobar.Plugins.Monitors.Common.Run
 import Xmobar.Plugins.Monitors.Cpu
-import Control.Monad.Reader
-import Data.IORef (newIORef)
 
 main :: IO ()
 main = do
-  cpuParams <- mkCpuArgs
-  defaultMain $ normalBench cpuParams
-    where
-      normalBench args = [ bgroup "Cpu Benchmarks" $ normalCpuBench args]
-
-runMonitor :: MConfig -> Monitor a -> IO a
-runMonitor config r = runReaderT r config
+  defaultMain =<< sequence [cpuBench]
 
 mkCpuArgs :: IO CpuArguments
-mkCpuArgs = getArguments ["-L","3","-H","50","--normal","green","--high","red", "-t", "Cpu: <total>%"]
-  
--- | The action which will be benchmarked
-cpuAction :: CpuArguments -> IO String
-cpuAction = runCpu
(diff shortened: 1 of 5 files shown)


jaor:
Thanks for the very informative exchange and the attention to detail, guys.  In my opinion, this patch puts together the best of both worlds (same amount of code changes, almost same performance, much simpler user interface), so i'm merging and i think we can discard #502, unless @psibi feels strongly for it.

PR #505, opened 2020-11-14
StdinReader: Improve exception handling

liskin:
This corrects a misleading comment "EOF check is necessary for certain
systems" which was added without complete understanding of the root
cause of #442. That issue was in fact caused by old xmobars not being
terminated on EOF, and is thus necessary on _all_ systems that rely on
EOF to terminate old xmobar before starting a new one.

Furthermore, this fixes another execution path that could lead to xmobar
not being terminated on EOF:

`echo -e '\xff' | xmobar -c '[Run StdinReader]' -t '%StdinReader%'`
would terminate the StdinReader thread upon catching the "invalid
argument (invalid byte sequence)" so there'd be no thread to detect the
subsequent EOF and xmobar would get stuck.

Additionally, I believe that terminating either the thread or the entire
xmobar upon receiving a single miscoded byte isn't desirable, as this
might be an intermittent issue and another input line can be perfectly
okay. Therefore I suggest that the original issue @psibi was trying to
fix by b7a3d6745817 is worked around by introducing a throttling delay
instead of terminating the thread, as I assume that exceptions other
than async and EOF are recoverable.

Fixes: b7a3d6745817 ("Avoid busy looping by not catching all exceptions")
Fixes: 68ac4d3ae6f3 ("Update stderr and the bar on receiving exception")
Fixes: ed0663aac942 ("Add EOF check before getLine operation from stdin")
Fixes: https://github.com/jaor/xmobar/issues/442
Related: https://github.com/jaor/xmobar/pull/439
Related: https://github.com/jaor/xmobar/pull/448

### DIFF
--- a/src/Xmobar/Plugins/StdinReader.hs
+++ b/src/Xmobar/Plugins/StdinReader.hs
@@ -27,7 +27,6 @@ import System.IO
 import System.IO.Error (isEOFError)
 import Xmobar.Run.Exec
 import Xmobar.X11.Actions (stripActions)
-import Control.Concurrent (threadDelay)
 import Control.Exception
 import Control.Monad (forever)
 
@@ -45,15 +44,11 @@ instance Exec StdinReader where
       -- there'd be a pileup of xmobars
       handler (fromException -> Just e) | isEOFError e = exitImmediately ExitSuccess
       -- any other exception, like "invalid argument (invalid byte sequence)",
-      -- is logged to both stderr and the bar itself, throttled to avoid
-      -- excessive CPU usage whenever someone pipes garbage into xmobar, and
-      -- then discarded without terminating, so a single charset error doesn't
-      -- break the entire xmobar
+      -- is logged to both stderr and the bar itself
       handler e = do
         let errorMessage = "xmobar: Received exception " <> show e
         hPutStrLn stderr errorMessage
         cb $ stripActions errorMessage
-        threadDelay 1000000
 
 escape :: StdinReader -> String -> String
 escape StdinReader = stripActions
--- a/src/Xmobar/Plugins/StdinReader.hs
+++ b/src/Xmobar/Plugins/StdinReader.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE ViewPatterns #-}
+
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  Plugins.StdinReader
@@ -22,30 +24,36 @@ import Prelude
 import System.Posix.Process
 import System.Exit
 import System.IO
+import System.IO.Error (isEOFError)
 import Xmobar.Run.Exec
 import Xmobar.X11.Actions (stripActions)
-import Xmobar.System.Utils (onSomeException)
-import Control.Monad (when)
+import Control.Concurrent (threadDelay)
+import Control.Exception
+import Control.Monad (forever)
 
 data StdinReader = StdinReader | UnsafeStdinReader
   deriving (Read, Show)
 
 instance Exec StdinReader where
-  start stdinReader cb = do
-    -- The EOF check is necessary for certain systems
-    -- More details here https://github.com/jaor/xmobar/issues/442
-    eof <- isEOF
-    when eof $
-      do hPrint stderr "xmobar: eof at an early stage"
-         exitImmediately ExitSuccess
-    s <-
-      getLine `onSomeException`
-      (\e -> do
-         let errorMessage = "xmobar: Received exception " <> show e
-         hPrint stderr errorMessage
-         cb errorMessage)
-    cb $ escape stdinReader s
-    start stdinReader cb
+  start stdinReader cb = forever $ (cb . escape stdinReader =<< getLine) `catch` handler
+    where
+      -- rethrow async exceptions like ThreadKilled, etc.
+      handler (fromException -> Just e) = throwIO (e :: SomeAsyncException)
+      -- XMonad.Hooks.DynamicLog.statusBar starts new xmobar on every xmonad
+      -- reload and the old xmobar is only signalled to exit via the pipe
+      -- being closed, so we must unconditionally terminate on EOF, otherwise
+      -- there'd be a pileup of xmobars
+      handler (fromException -> Just e) | isEOFError e = exitImmediately ExitSuccess
+      -- any other exception, like "invalid argument (invalid byte sequence)",
+      -- is logged to both stderr and the bar itself, throttled to avoid
+      -- excessive CPU usage whenever someone pipes garbage into xmobar, and
+      -- then discarded without terminating, so a single charset error doesn't
+      -- break the entire xmobar
+      handler e = do
+        let errorMessage = "xmobar: Received exception " <> show e
+        hPutStrLn stderr errorMessage
+        cb $ stripActions errorMessage
+        threadDelay 1000000
 
 escape :: StdinReader -> String -> String
 escape StdinReader = stripActions
--- a/src/Xmobar/System/Utils.hs
+++ b/src/Xmobar/System/Utils.hs
@@ -20,7 +20,6 @@
 module Xmobar.System.Utils
   ( expandHome
   , changeLoop
-  , onSomeException
   , safeIndex
   ) where
 
@@ -31,7 +30,6 @@ import Data.Maybe (fromMaybe)
 
 import System.Environment
 import System.FilePath
-import Control.Exception
 
 expandHome :: FilePath -> IO FilePath
 expandHome ('~':'/':path) = fmap (</> path) (getEnv "HOME")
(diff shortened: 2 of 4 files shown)


src/Xmobar/Plugins/StdinReader.hs
```diff
@@ -22,30 +24,36 @@ import Prelude
...
+      -- reload and the old xmobar is only signalled to exit via the pipe
+      -- being closed, so we must unconditionally terminate on EOF, otherwise
+      -- there'd be a pileup of xmobars
+      handler (fromException -> Just e) | isEOFError e = exitImmediately ExitSuccess
+      -- any other exception, like "invalid argument (invalid byte sequence)",
+      -- is logged to both stderr and the bar itself, throttled to avoid
+      -- excessive CPU usage whenever someone pipes garbage into xmobar, and
+      -- then discarded without terminating, so a single charset error doesn't
+      -- break the entire xmobar
+      handler e = do
```
psibi, line 52: I think if we receive an exception like `"invalid argument (invalid byte sequence)"`, it's a bug somewhere. I remember fixing such a big recently on both xmobar and xmonad-contrib side. So, I'm worried that this handling might bury bugs.

psibi (commented):
LGTM apart from a single comment.

src/Xmobar/Plugins/StdinReader.hs
liskin: Well the error message will still be logged to stderr and will be visible on screen for at least a second, so I wouldn't call that "burying". But I'm not familiar with the specific bug you're talking about. Would there be another line that wouldn't trigger this handler and hide the error message?

psibi: >  so I wouldn't call that "burying"

Usually I won't go and check `stderr` or `xsession.errors` unless something stops working. But I do think this improves the current situation! :-)

>  But I'm not familiar with the specific bug you're talking about.

* https://github.com/xmonad/xmonad-contrib/pull/334
* https://github.com/xmonad/xmonad-contrib/pull/346

liskin: > Usually I won't go and check stderr or xsession.errors unless something stops working. But I do think this improves the current situation! :-)

Well, what we could do is to show the error but also keep reading from stdin (to be able to detect EOF an terminate) and just discard whatever we read. Or a mix of both: always show the last error, if any, and append the line from stdin (if decodable, obviously) to it. I'm not sure if all this is worth it but it can certainly be done. :-)

jaor (approved):
this looks reasonable (and correct) to me, so i'm fine merging if
there are no further comments.  I think we need a rebase, though: @liskin would
you mind doing it please?

src/Xmobar/Plugins/StdinReader.hs
jaor: i'm not sure either.  is that something that we could add later?

src/Xmobar/Plugins/StdinReader.hs
```diff
@@ -22,30 +24,36 @@ import Prelude
...
+        let errorMessage = "xmobar: Received exception " <> show e
+        hPutStrLn stderr errorMessage
+        cb $ stripActions errorMessage
+        threadDelay 1000000
```
psibi, line 56: Won't the `threadDelay` just delay the next `getLine` ? Won't this cause issue when multiple subsequent lines cause the exception and each of them just make it to stall one second... Probably it's okay to have the current behavior itself.

liskin:
> I think we need a rebase

There were no conflicts so I'm really confused why I had to do this... did it nevertheless.

src/Xmobar/Plugins/StdinReader.hs
liskin: Yes, it will delay and it will cause issues in that case.

Unfortunately there isn't a good solution here. We're trying to detect and exit on EOF, and at the same time we're trying not to consume 100% CPU when someone pipes /dev/urandom into it, and there's no way to do both perfectly.

Now if you have a way to reproduce the 100% CPU bug without piping total garbage into it, I might optimize it for that scenario. :-)

psibi: > and at the same time we're trying not to consume 100% CPU when someone pipes /dev/urandom into it, and there's no way to do both perfectl

I personally feel we shouldn't try to handle this situation and should just avoid `threadDelay`. We should just read the next line, hoping for the best. :-)

liskin: Okay, that would be fine by me, but I'd still love to know the exact cause of #438 that you were trying to fix with b7a3d674581720bfb63bf73cae8368ebbad81004. I don't like introducing regressions without understanding the tradeoffs.

I still don't see how xmobar could eat 100% CPU with just a few lines on input. If that's possible, I'd love to have a reproducer.

psibi: > but I'd still love to know the exact cause of #438 that you were trying to fix with b7a3d67

Sure, this was the cause of my issue: https://github.com/jaor/xmobar/issues/438#issuecomment-623378023

It happened because my xmonad + xmobar is being managed via Nix in an Ubuntu machine.

> I still don't see how xmobar could eat 100% CPU with just a few lines on input. If that's possible, I'd love to have a reproducer.

I think the reproduction step is pretty involved as it requires a Nix package manager on a non NixOS distribution. You would have to do the following things:
* Undo my patch: https://github.com/jaor/xmobar/pull/439/files
* Undo my xmonad-contrib patches where I fix the encoding related issues: https://github.com/xmonad/xmonad-contrib/pull/334, https://github.com/xmonad/xmonad-contrib/pull/346
* Make sure in your Linux distribution the `LOCALE_ARCHIVE` is not set and that both your xmonad and xmobar is managed via Nix package manager.
* Open your browser and visit a page with some unicode characters. This seemed to work for me at that time: https://www.manoramaonline.com/news/latest-news.html
* See your xmobar getting CPU throttled now

liskin: Well yeah but what is the exact input that xmonad sends to xmobar? Does it send just one line or does it keep sending many lines in a loop? And if it does send them in a loop, why?

psibi: > Well yeah but what is the exact input that xmonad sends to xmobar? Does it send just one line or does it keep sending many lines in a loop? And if it does send them in a loop, why?

Sorry, don't have much information about that!

jaor:
On Wed, Nov 18 2020, Tomáš Janoušek wrote:

>     I think we need a rebase
>    
> There were no conflicts so I'm really confused why I had to do
> this... did it nevertheless.

because i merged your other PR first.

liskin:
> because i merged your other PR first

Yeah, but there were no conflicts. 😕

src/Xmobar/Plugins/StdinReader.hs
liskin: :-(

liskin: Anyway, if you're okay with removing the threadDelay, I can probably remove it, as I don't think anyone else encountered the issue where any throttling would be needed. If you or anyone else encounter it again, feel free to wake me up in the middle of the night to debug it. :-))

src/Xmobar/Plugins/StdinReader.hs
liskin: Anything can be added later. :-)

liskin:
I dropped the throttling, updated the commit messages with more details, and fixed the test.

jaor:
On Wed, Nov 18 2020, Tomáš Janoušek wrote:

> Anything can be added later. :-)

yes, that's true.  i don't use this monitor, and never had any problem
with it in the past when i did, nor have been able to force them. so i'm
trusting you guys when you tell me it's fine to merge... it'd be nice to
hear @slotThe 's opinion too (if i'm not misremembering, he had issues
with the original change back in the day).

src/Xmobar/Plugins/StdinReader.hs
psibi: Thanks for removing the `threadDelay`! 

>  If you or anyone else encounter it again, feel free to wake me up in the middle of the night to debug it. :-))

:-)

slotThe:
On Wed, Nov 18 2020 12:35, jao wrote:
> yes, that's true.  i don't use this monitor, and never had any problem
> with it in the past when i did, nor have been able to force them. so i'm
> trusting you guys when you tell me it's fine to merge... it'd be nice to
> hear @slotThe 's opinion too (if i'm not misremembering, he had issues
> with the original change back in the day).

Yeah, we found out in [1] that this is due to some weird setup I have
with regards to compiling xmonad and xmobar.  In either case, I just
tested this change and it doesn't give me any problems.  Code-wise it
looks good as well, so no objections here!

[1] https://github.com/xmonad/xmonad-contrib/pull/400

jaor:
excellent, @slotThe.  thanks again to all of you for your thorough explorations and analyses, it is really rewarding as a maintainer.  merged!

PR #509, opened 2020-11-25
Use tripple backticks for Haskell code

### LINKED ISSUE #508: Use tripple backticks in readme
A thing I've noticed while working on a new monitor:

In the `readme.md` file we are using indentation to mark things as code.
Sadly, this doesn't add any syntax highlighting to the examples, but
because the configuration language is essentially a subset of Haskell, I
think it would be a net-win if we (using the tripple backtick syntax)
marked everything as Haskell code.

I would do this myself, but I can see (well, it's more of a guess) that
the readme is generated from `org-mode`, and I can't find a `readme.org`
file anywhere!  It's possible to do this by using e.g. [1] to export the
file to "GitHub flavoured markdown".

[1] https://github.com/larstvei/ox-gfm

slotThe:
Closes #508 

I also made the indentation of some other code a little bit more regular, removing a few inconcistencies there.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -182,11 +182,15 @@ Otherwise, you'll need to install them yourself.
   use the `xft:` prefix in the `font` configuration option. For
   instance:
 
+    ``` haskell
         font = "xft:Times New Roman-10:italic"
+    ```
 
   Or to have fallback fonts, just separate them by commas:
 
+    ``` haskell
         font = "xft:Open Sans:size=9,WenQuanYi Zen Hei:size=9"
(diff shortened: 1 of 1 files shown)


jaor (approved):

jaor:
excellent.  many thanks for taking the time!

slotThe:
On Wed, Nov 25 2020 09:07, jao wrote:
> excellent.  many thanks for taking the time!

Ah no problem; it was quite meditative in a way :)

PR #511, opened 2020-11-29
New Plugin: NotmuchMail

slotThe:
New plugin that checks for new mail, provided that this mail is indexed
by `notmuch`.

As mail that was tagged is moved from the `new` directory to `cur`, the
`inotify` solution that the current `Mail` plugin (as well as its
variants) uses won't work for such mail.

### DIFF
--- a/src/Xmobar/Plugins/NotmuchMail.hs
+++ b/src/Xmobar/Plugins/NotmuchMail.hs
@@ -18,8 +18,8 @@
 --
 -- As mail that was tagged is moved from the @new@ directory to @cur@,
 -- the @inotify@ solution that he mail 'Mail' plugin (and its variants)
--- uses won't work for such mail.  Hence we have to resort to refreshing
--- every N time intervals.
+-- uses won't work for such mail.  Hence, we have to resort to a
+-- refresh-based monitor.
 --
 -- Note that, in the `notmuch` spirit, this plugin checks for new
 -- threads and not new individual messages.  For convenience, the
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@ _New features_
   - New command line option `--add-font` (Ivan Brennan)
   - New monitor `MPDX` that extends `MPD` with the ability of having a
     custom alias.  Useful for connecting with multiple servers.
+  - New plugin `NotmuchMail` to monitor mail indexed by `notmuch`.
 
 _Bug fixes_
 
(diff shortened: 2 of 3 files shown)


src/Xmobar/Plugins/NotmuchMail.hs
```diff
@@ -0,0 +1,146 @@
...
+>
+>     -- All unread mail that has @[Haskell-Cafe]@ in the subject line.
+>   , MailItem "C:" "" "subject:[Haskell-Cafe]"
+>
+>     -- All unread mail that's tagged as @lists@, but not @haskell@.
+>   , MailItem "H:" "" "tag:lists and not tag:haskell"
+>   ]
+>   600                   -- update every 60 seconds
+
+-}
```
jaor, line 93: thanks for the excellent documentation.  a thought here: the above is user-level
information, unrelated to the implementation.  users of xmobar that simply
install it and know nothing about haskell won't (and shouldn't, imo) look for
API-level docs about haskell modules.  i would put the above detailed
manual in the readme, and here the summary you put in the readme.
of course things would be different if we had an automatic way of generating
user-level docs from these docstrings, but we don't (i don't think the
api docs qualify as that), and, unfortunately, we haven't been as careful
as you are here documenting our modules (thanks again!), so even if we had
the resulting manual wouldn't be much to look at :).

jaor (commented):
this looks very nice.  please see below a comment on docs
(i'm happy merging as it is, though)

slotThe:
On Sun, Nov 29 2020 08:38, jao wrote:
> thanks for the excellent documentation.  a thought here: the above is user-level
> information, unrelated to the implementation.  users of xmobar that simply
> install it and know nothing about haskell won't (and shouldn't, imo) look for
> API-level docs about haskell modules.  i would put the above detailed
> manual in the readme, and here the summary you put in the readme.
> of course things would be different if we had an automatic way of generating
> user-level docs from these docstrings, but we don't (i don't think the
> api docs qualify as that), and, unfortunately, we haven't been as careful
> as you are here documenting our modules (thanks again!), so even if we had
> the resulting manual wouldn't be much to look at :).

Thanks for the comment!  You are right of course, I think I was still in
"xmonad-mode" when I wrote this... :)

I have moved the relevant stuff to the readme and shortened the module
documentation to just the implementation specific things, as well as
some gotcha's that I thought were still worth mentioning.

jaor (approved):

jaor:
merged, thanks!

PR #516, opened 2020-12-15
Subdivide doc/plugins.org

slotThe:
This implements the idea

> idea for the future: it'd might be useful to categorize our monitors
> in groups with their own section headings, e.g. "Battery monitors",
> "Email".... and so on.

from #514

It also recommends (visually, by simply listing them first)
property-based logging over using pipes, due to the various issues with
pipe-based solutions.

### DIFF
--- a/doc/plugins.org
+++ b/doc/plugins.org
@@ -1264,158 +1264,178 @@ and not new individual messages.
 Listed below are ways to interface xmobar with your window manager of
 choice.
 
-** =XMonadLog=
+** Property-based Logging
+*** =XMonadLog=
 
-- Aliases to XMonadLog
+ - Aliases to XMonadLog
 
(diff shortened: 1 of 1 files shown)


jaor (approved):
this is excellent, thanks!.  only a minor tweak: i would expect Swap to be in Memory rather than Disk Monitors... is it just me?

also, i've been preferring property monitors over stdin and pipes for years, so i can only applaud your decision :)

slotThe:
On Tue, Dec 15 2020 10:08, jao wrote:
> this is excellent, thanks!.  only a minor tweak: i would expect Swap
> to be in Memory rather than Disk Monitors... is it just me?

Ah yeah good point; I wasn't too sure on some of these assignments and
don't have any strong opinions on where Swap goes.  I just pushed a
change that puts it where Memory is as well

jaor:
perfect. it's in!

PR #521, opened 2021-01-15
Extract compiling, writing a plugin to their own files

slotThe:
This implements two suggestions of yours from #514, namely

  1. Putting all the information related to compiling its their own
     `.org` file
  2. Putting all the information related to writing your own plugins
     (I'm not sure about the name of the file here, but it fits, I
     suppose :) in its own `.org` file


jaor:
looks good, merged!

PR #522, opened 2021-01-28
Add plugins.org to readme

slotThe:
Quick, but important, one.  
The readme should contain _all_ of this kind of information


jaor:
indeed. merged!

PR #531, opened 2021-03-18
Fix typo `DRTSOPTS` -> `RTSOPTS`

avdv:
The cpp option `-DRTSOPTS` results in `RTSOPTS` being defined.


jaor:
merged, thanks a lot!

PR #533, opened 2021-03-30
Add FreeBSD support to Cpu plugin

Partmedia:
This pull request adds FreeBSD support to the Cpu plugin.

This patch moves CpuData to the top (outside of the new #ifdef) and provides a *cpuData* and *parseCpu* implementation for FreeBSD. This implementation was tested and works on FreeBSD, but someone should check that I haven't managed to break the Linux build.


jaor:
Works for me on linux, and the CI failure seems totally unrelated (bad
temper of hlint), so merging.  Thanks!

Issue #530, opened 2021-03-16
Full width next to vertical display


capitanu:
Take the example of a two monitor setup: left is horizontal and right is vertical.

If the xmobar on the left has full monitor width (1920 px), the vertical display on the right will only show all the applications in the lower half of the display. If I set the width of the left xmobar to 1919, everything works as expected.

liskin:
That looks like a bug in your window manager rather than xmobar. I vaguely remember fiddling with something related in xmonad around 2009/2010, but I've never had a problem since. What WM are you using?

capitanu:
Sorry for the late reply! I did not notice your reply.

I am using xmonad! Do you remember what fixed it for you? It's not a huge problem, but would be nice to fix!

liskin:
Yeah, I do remember: https://github.com/xmonad/xmonad-contrib/commit/82a0d30f31135dfd884fff6afa5f7ca989352382, but it's almost certainly just "vaguely related", there's no way this will help you.

Your problem sounds more like an issue with a border or something. Can you perhaps show what `xrandr --verbose` shows and what `xwininfo` shows when you point it at xmobar?

liskin:
Were you by any chance using `xmobar` with Static positioning instead of `-x`? Does https://github.com/jaor/xmobar/pull/538 fix your issue?

PR #538, opened 2021-04-26
Fix off-by-one in strut calculation for Static position

liskin:
Pixel intervals in _NET_WM_STRUT_PARTIAL are closed, not open [1], so xmobar of width 1920 at x=0 needs to be encoded as ⟨0, 1919⟩, not ⟨0, 1920). The code already does this for Top/Bottom positioning, but Static has been buggy for years.

[1]: https://specifications.freedesktop.org/wm-spec/1.5/ar01s05.html#idm46075117229424

Fixes: https://old.reddit.com/r/xmonad/comments/mynegr/xmobar_dual_monitor_bug/
Fixes: https://github.com/jaor/xmobar/issues/530


jaor:
pushed, thanks!

PR #544, opened 2021-05-19
Add k10temp monitor

sam-kirby:
The existing support for the coretemp kernel driver only works with
Intel CPUs.

This commit extends support for temperature monitoring to AMD CPUs.

k10temp is a kernel driver for monitoring the temperature of AMD
processors. It supports everything from AMD's 10h (Opteron/Athlon)
family of processors to the latest 19h (Zen 3) family.

Reference: https://www.kernel.org/doc/html/latest/hwmon/k10temp.html

The meaning of the various temperatures made available has changed over
the years and only `Tctl` is available on processors prior to the 17h
family.

Labels for these temperatures are present but as Tctl and Tdie do not
contain a number I could not find a way to use these as
`checkedDataRetrieval` expects an integer label.

It is a PCI device and so an address needs to be supplied as part of the
configuration.

Example configuration:
`Run K10Temp "0000:00:18.3" ["--template", "T: <Tdie>C | <Tccd1>C"] 60`

### DIFF
--- a/doc/plugins.org
+++ b/doc/plugins.org
@@ -455,6 +455,39 @@ something like:
                         "-l", "green", "-n", "yellow", "-h", "red",
                         "--", "--mintemp", "20", "--maxtemp", "100"] 50
    #+end_src
+
+*** =K10Temp Slot Args RefreshRate=
+
+ - Aliases to =k10temp=
+
+ - Slot: The PCI slot address of the k10temp device
+
+ - Args: default monitor arguments
+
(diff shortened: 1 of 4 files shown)


jaor:
this looks good!  would you have the bandwidth to add a bit of
documentation in a new section here:
https://github.com/jaor/xmobar/blob/feat%2Fk10temp/doc/plugins.org#L619
?

thanks a lot!

jaor:
excellent. merged!

PR #548, opened 2021-05-24
NotmuchMail: Properly parse the type for non-Haskell based configurations

### LINKED ISSUE #547: Can't get NotmuchMail Runnable to parse in xmobarrc
I am trying to set up the new `NotmuchMail` plugin, however I must either be including a syntax error or something else is amiss.

My xmobarrc looks like this:
```
Config { 

   -- appearance
     font =         "xft:Bitstream Vera Sans Mono:size=7:bold:antialias=true"
   , bgColor =      "black"
   , fgColor =      "#646464"
   , position =     Static { xpos = 0, ypos = 16, width = 1920, height = 16 }
   , border =       BottomB
   , borderColor =  "#646464"

   -- layout
   , sepChar =  "%"   -- delineator between plugin names and straight text
   , alignSep = "}{"  -- separator between left-right alignment
   , template = "}{ %battery% | %memory% | %multicpu%"

   -- general behavior
   , lowerOnStart =     True    -- send to bottom of window stack on start
   , hideOnStart =      False   -- start with window unmapped (hidden)
   , allDesktops =      True    -- show on all desktops
   , overrideRedirect = True    -- set the Override Redirect flag (Xlib)
   , pickBroadest =     False   -- choose widest display (multi-monitor)
   , persistent =       True    -- enable/disable hiding (True = disabled)

   -- plugins
   --   Numbers can be automatically colored according to their value. xmobar
   --   decides color based on a three-tier/two-cutoff system, controlled by
   --   command options:
   --     --Low sets the low cutoff
   --     --High sets the high cutoff
   --
   --     --low sets the color below --Low cutoff
   --     --normal sets the color between --Low and --High cutoffs
   --     --High sets the color above --High cutoff
   --
   --   The --template option controls how the plugin is displayed. Text
   --   color can be set by enclosing in <fc></fc> tags. For more details
   --   see http://projects.haskell.org/xmobar/#system-monitor-plugins.
   , commands = 

        -- network activity monitor (dynamic interface resolution)
        [ Run DynNetwork     [ "--template" , "<dev>: <tx>kB/s|<rx>kB/s"
                             , "--Low"      , "1000"       -- units: B/s
                             , "--High"     , "5000"       -- units: B/s
                             , "--low"      , "darkgreen"
                             , "--normal"   , "darkorange"
                             , "--high"     , "darkred"
                             ] 10

        -- cpu activity monitor
        , Run MultiCpu       [ "--template" , "Cpu: <autototal>%"
                             , "--Low"      , "50"         -- units: %
                             , "--High"     , "85"         -- units: %
                             , "--low"      , "darkgreen"
                             , "--normal"   , "darkorange"
                             , "--high"     , "darkred"
                             , "--ppad"     , "2"
                             ] 10

        -- memory usage monitor
        , Run Memory         [ "--template" ,"Mem: <usedratio>%"
                             , "--Low"      , "20"        -- units: %
                             , "--High"     , "90"        -- units: %
                             , "--low"      , "darkgreen"
                             , "--normal"   , "darkorange"
                             , "--high"     , "darkred"
                             ] 10

        -- battery monitor
        , Run Battery        [ "--template" , "Batt: <acstatus>"
                             , "--Low"      , "10"        -- units: %
                             , "--High"     , "80"        -- units: %
                             , "--low"      , "darkred"
                             , "--normal"   , "darkorange"
                             , "--high"     , "darkgreen"

                             , "--" -- battery specific options
                                       -- discharging status
                                       , "-o"	, "<left>% (<timeleft>)"
                                       -- AC "on" status
                                       , "-O"	, "Charging (<left>%)"
                                       -- charged status
                                       , "-i"	, "Charged"
                             ] 50

        , Run NotmuchMail "mail" [MailItem "unread" "" ""] 3000

        ]
   }

```

However this fails with
```
Invalid configuration file: "Config" (line 88, column 5):
error reading the commands: this usually means that a command could not
be parsed.
The error could be located at the begining of the command
which follows the offending one. field: [ Run DynNetwork     [ "--template" , "<dev>: <tx>kB/s|<rx>kB/s"
[... the rest of the rcfile]
```
where line 88 is the closing `}`.

The `NotmuchMail` plugin should be available even without the `all_extensions` flag, at least according to the cabal file.

One thing I noticed however, is that the type constructor is not in [this list](https://github.com/jaor/xmobar/blob/82461417abbcd2e03e4329d68df4a15866885dd2/src/Xmobar/Run/Types.hs#L55), could that have something to do with it?

slotThe:
Fixes #547 

1.  Add NotmuchMail as a runnable type 

This is needed for the plugin to parse properly in non-Haskell based 
configurations.

2.  NotmuchMail: Manually implement Read instance

The automatically derived read instance expects the type to be given in
record syntax; this is not what most users want.  In order to simply
specify the type via

    Run NotmuchMail "mail" [MailItem "name" "" "tag:unread"] 3000

we have to write our own Read instance.

3.  NotmuchMail: Update documentation

Ever since d51b9d5 we can't specify a
MailItem via the record syntax anymore.  This could conceivably be added
back in by being part of the Read instance, but that would make it more
complicated for perhaps little gain.

### DIFF
--- a/doc/plugins.org
+++ b/doc/plugins.org
@@ -911,18 +911,6 @@ something like:
      ]
    #+end_src
 
-   or, using explicit record syntax:
-
-   #+begin_src haskell
-     [ MailItem
-         { name    = "name"
-         , address = "address"
-         , query   = "query"
-         }
-       ...
-     ]
-   #+end_src
-
    where
 
    - =name= is what gets printed in the status bar before the number of
--- a/src/Xmobar/Plugins/NotmuchMail.hs
+++ b/src/Xmobar/Plugins/NotmuchMail.hs
@@ -39,6 +39,7 @@ import Control.Concurrent.Async (mapConcurrently)
 import Data.Maybe (catMaybes)
 import System.Exit (ExitCode(ExitSuccess))
 import System.Process (readProcessWithExitCode)
+import Text.Read (Lexeme(Ident), ReadPrec, lexP, parens, prec, readPrec, reset)
 
 
 -- | A 'MailItem' is a name, an address, and a query to give to @notmuch@.
@@ -48,7 +49,13 @@ data MailItem = MailItem
                        --   the empty string to query all indexed mail instead
   , query   :: String  -- ^ Query to give to @notmuch search@
   }
-  deriving (Read, Show)
+  deriving (Show)
+
+instance Read MailItem where
+  readPrec :: ReadPrec MailItem
+  readPrec = parens . prec 11 $ do
+    Ident "MailItem" <- lexP
+    MailItem <$> reset readPrec <*> reset readPrec <*> reset readPrec
 
 -- | A full mail configuration.
 data NotmuchMail = NotmuchMail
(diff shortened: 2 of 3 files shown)


jaor (approved):
this looks good, thanks.  and i tend to agree the additional complication
is not really worth it.

PR #550, opened 2021-06-18
Fix MultiCoreTemp --hwmon-path argument docs

ch1bo:
The document seems to be off. in the code `hwmon-path` is parsed from the `Args`


jaor:
thanks!

PR #551, opened 2021-06-18
Add Gentoo to installation method. 

187Qrly:
Xmobar is also avalible under gentoo, so why don't make a installation method there to in the readme?


jaor:
why not indeed.  thanks!

PR #552, opened 2021-06-29
Add Kraken plugin

amir:
I understand this might not be a Plugin suited to be part of the official project but I thought other people might find it useful too. It can be configured like:

```
Run Kraken ["ADA/EUR", "XTZ/EUR", "NANO/EUR"] "kraken"
```

Which then will produce the following and updates it via Websockets.

2021-06-29-224324_3840x2160_scrot

Disclaimer: I'm in no way affiliated with Kraken, I just used their API because it's the only free and reliable one I could find that doesn't require an account.

### DIFF
--- a/src/Xmobar/Plugins/Kraken.hs
+++ b/src/Xmobar/Plugins/Kraken.hs
@@ -21,7 +21,7 @@ import qualified Data.Map as Map
 import qualified Data.Vector as V
 
 data Kraken = Kraken [String] String
-  deriving (Read,Show)
+  deriving (Read, Show)
 
 instance Exec Kraken where
   alias (Kraken _ a) = a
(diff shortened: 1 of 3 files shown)


jaor:
hi @amir and thanks for your contribution.  the code per se looks good enough to be merged.  i must confess i know nearly nothing about kranken and their business, and i'm thus unsure as to whether we should merge this or not.  in principle, i see it's a public service that doesn't even require signing up, so i don't see anything wrong with it prima facie... unless we're talking here about some kind of evil corporation, for some definition of evil :) opinions anyone?

amir:
Hi @jaor, thanks for your input. That's a valid and understandable concern to which, sadly, I don't have an answer.

jaor:
@amir since nobody seems to be specially troubled and the plugin is opt-in anyway, if you rebase this PR on top of current master (and the CI tests pass), i think we can merge it.  thanks!

amir:
@jaor thanks, I just rebased the branch.

PR #554, opened 2021-07-06
Filter filename when executing Haskell-based config

slotThe:
We are now—in case the user specified a Haskell file as their xmobar
configuration—threading the command line arguments that xmobar receives
to the relevant execv() call.  However, we simply shove in all arguments
originally given to xmobar, including the path to the configuration
file.  As main is now defined within that very file, this seems
unneccessary.

By filtering out that part of the arguments, the pattern that a lot of
users seem to follow for easy setting of certain options becomes a
little bit nicer.  For example:

    main :: IO ()
    main = getArgs >>= \case
      ["-x", n, _] -> xmobar . config $ read n
      _            -> xmobar $ config 0

becomes

    main :: IO ()
    main = getArgs >>= \case
      ["-x", n] -> xmobar . config $ read n
      _         -> xmobar $ config 0

Related: https://github.com/jaor/xmobar/pull/553

### DIFF
--- a/src/Xmobar/App/Main.hs
+++ b/src/Xmobar/App/Main.hs
@@ -106,5 +106,5 @@ xmobarMain = do
     Just p -> do r <- readConfig defaultConfig p
                  case r of
                    Left e ->
-                     buildLaunch args (verboseFlag flags) (recompileFlag flags) p e
+                     buildLaunch (filter (/= p) args) (verboseFlag flags) (recompileFlag flags) p e
                    Right (c, defs) -> doOpts c flags >>= xmobar' defs


jaor:
This looks like a good idea to me.  But you made me think: have we
potentially "broken" user code with #553 or this one? Broken in the
sense that they'll need to modify it when they update... just to
mention it in the changelog at least.

slotThe:
I don't think we have broken any user code.  There are a few ways to use xmobar: 

  - People with a normal config file: We are not compiling, nor execing anything here, so they should be completely unaffected.
  - People with project-based configurations: Since users have to define their own `main` function, `xmobarMain` never gets called and so there is no handling of command line arguments whatsoever and everything has to be self-defined anyways. 
  - People with Haskell-based configuration files, but no projects to compile these directly into xmobar: as far as I can tell, it has just been impossible to use xmobar's command line arguments up to #553; everything that you gave xmobar was just swallowed up.

Though perhaps mentioning somewhere that this is now in fact possible would be worth it

jaor:
thanks, i think you're right: i've merged and added a note to the changelog.

Issue #555, opened 2021-07-08
Bar on bottom is broken.


exorcist365:
When the position is set to Bottom, and the height is more than 20 the bar doesn't render the rest, like so: https://0x0.st/-O-6.png

lua-latex:
This is issue i agree

PR #556, opened 2021-07-11
Add TopH and BottomH for only controlling height of the window.

exorcist365:
Yeah, the title says it all, I have been using it for the past week and had no issues except for #555 but that happens when the position is set to static {} as well.

### DIFF
--- a/src/Xmobar/X11/Window.hs
+++ b/src/Xmobar/X11/Window.hs
@@ -80,11 +80,11 @@ setPosition c p rs ht =
   case p' of
     Top -> Rectangle rx ry rw h
     TopP l r -> Rectangle (rx + fi l) ry (rw - fi l - fi r) h
-    TopH f  -> Rectangle rx ry rw (fi f)
+    TopH ch -> Rectangle rx ry rw (fi ch)
     TopW a i -> Rectangle (ax a i) ry (nw i) h
     TopSize a i ch -> Rectangle (ax a i) ry (nw i) (mh ch)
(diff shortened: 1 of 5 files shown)


jaor:
this looks good, thank you.  could you please mention the new options
in the doc?

exorcist365:
Yeah, sure.

src/Xmobar/X11/Window.hs
```diff
@@ -80,9 +80,11 @@ setPosition c p rs ht =
   case p' of
     Top -> Rectangle rx ry rw h
     TopP l r -> Rectangle (rx + fi l) ry (rw - fi l - fi r) h
+    TopH ch -> Rectangle rx ry rw (fi ch)
```
liskin, line 83: I wonder why this uses `ch` directly but `TopSize` uses `mh ch`. Is that useful in practice?

exorcist365: Do I make it use mh ch or make TopSize use just ch?

jaor: in theory, mh ch would be more correct, although i think it should never happen that they differ in this case.  it think it's better if we don't change TopSize.

btw, @exorcist365, thanks for the docs! would you mind squashing all your changes to a single commit (now it looks like a collection of 4)?

exorcist365: Yeah, done.

jaor (commented):
merged, thanks

Issue #567, opened 2021-09-14
HandleReader only ever says `Updating...`


guygastineau:
https://github.com/jaor/xmobar/blob/3a5004e39e6dc42f2691012b4513a210202311b2/src/Xmobar/Plugins/HandleReader.hs#L43

liskin:
Possibly a locale/encoding issue? HandleReader lacks exception handling, so I'm not sure if it'd print the exception anywhere, but probably worth checking the stderr anyway.

guygastineau:
I will post some of my config below.  Is anyone else using this?  I am trying to move toward completely embedding xmobar in my xmonad builds.  This is the last obstacle.

I initially tried it exactly as shown in the haddock example, but I couldn't get the posix processes to terminate, so I switched to `forkIO`.  The behavior of the `HandleReader` is the same both ways (but `forkIO` seems the safer way to do it).

Here is what I export from my `Top.hs` config file for a top bar.

``` haskell
runTop :: IO (ThreadId, Handle)
runTop = do
  (r, w) <- createPipe
  bar <- forkIO $ xmobar conf
    { commands = Run (HandleReader r "XMonadLog") : commands conf }
  return (bar, w)
```

I use it from my xmonad config like this:

``` haskell

main :: IO ()
main = do
  (_, logHandle) <- runTop
  run $ conf logHandle
  where
    run :: (LayoutClass l Window, Read (l Window)) => XConfig l -> IO ()
    run = xmonad . ewmh . myAdditionalKeys

    conf topBar = def
      { modMask  = mod4Mask
      , terminal = myTerminal

      -- style and color
      , borderWidth        = myBorderWidth
      , normalBorderColor  = myNormalBorderColor
      , focusedBorderColor = myFocusedBorderColor

      -- hooks
      , logHook         = myLogHook topBar
      , manageHook      = manageDocks <+> manageHook def
      , layoutHook      = avoidStruts myLayoutHook
      , handleEventHook = myHandleEventHook
      -- startup
      , startupHook     = setWMName "LG3D"
      }
```

The `myLogHook` just wraps the default and uses the pretty printing logger.

guygastineau:
@liskin Thank you for the suggestion.  I thought this could be an issue.  I didn't get anything in `~/.xmonad/xmonad.errors`  I will check my TTY too.

jsoo1:
Funny enough, I just found the same thing.

liskin:
[...]

PR #568, opened 2021-09-14
Add the QueueReader plugin.

guygastineau:
* A queue reader for xmobar using `TQueue a` from `stm`.
    This is a flexible and performat solution for sharing
    data between arbitrary haskell and xmobar.
  * I am not sure if I did the haddocks correctly.

---

Should there be an safe and unsafe version?  If `stripActions` were exposed, then users could make their own safe version trivially.  Let me know.

I ran `hlint`.  No errors in the whole codebase.  Awesome maintenance job you all!

---

I didn't make any tests.  Tests seem hard and unnecessary for this.  The plugin is very simple, so it practically feels tautological, but it feels arrogant to say this.

Let me know if I should make any improvements.

### DIFF
--- a/doc/plugins.org
+++ b/doc/plugins.org
@@ -1436,6 +1436,50 @@ choice.
      hPutStr writeHandle "Hello World"
    #+end_src
 
+** Software Transactional Memory
+
+When invoking xmobar from other Haskell code it can be easier and
+more performant to use shared memory.  The following plugins leverage
+=Control.Concurrent.STM= to realize these gains for xmobar.
+
+*** =QueueReader (TQueue a) (a -> String) String=
+
+ - Display data from a Haskell =TQueue a=.
+
+ - This plugin is only useful if you are running xmobar from another
+   haskell program like xmonad.
+
+ - You should make an =IO= safe =TQueue a= with =Control.Concurrent.STM.newTQueueIO=.
+   Write to it from the user code with =writeTQueue=, and read with =readTQueue=.
+   A common use is to overwite =ppOutput= from =XMonad.Hooks.DynamicLog= as shown
+   below.
+
+   #+begin_src haskell
+     main :: IO ()
+     main = do
+       q <- STM.newTQueueIO @String
+       bar <- forkIO $ xmobar myConf
+         { commands = Run (QueueReader q id "XMonadLog") : commands myConf }
+       xmonad $ def { logHook = logWorkspacesToQueue q }
+
+     logWorkspacesToQueue :: STM.TQueue String -> X ()
+     logWorkspacesToQueue q =
+       dynamicLogWithPP def { ppOutput = STM.atomically . STM.writeTQueue q }
+       where
+         -- Manage the PrettyPrinting configuration here.
+         ppLayout' :: String -> String
+         ppLayout' "Spacing Tall"        = xpm "layout-spacing-tall"
+         ppLayout' "Spacing Mirror Tall" = xpm "layout-spacing-mirror"
+         ppLayout' "Spacing Full"        = xpm "layout-full"
+         ppLayout' x = x
+
+         icon :: String -> String
+         icon path = "<icon=" ++ path ++ "/>"
+
+         xpm :: String -> String
+         xpm = icon . (++ ".xpm")
+   #+end_src
+
 * Executing External Commands
 
 In order to execute an external command you can either write the command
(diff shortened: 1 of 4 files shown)


guygastineau:
Inspired by #567

jaor (approved):

jaor:
Thanks Guy, this looks good... i think exposing in the library
stripActions is a good idea; having also a safe version would be useful
for people not using xmobar as a library, but, iiuc, they won't be using
this module at all!  would it be possible to have some docs for it in
doc/plugins.org (somewhere near
https://github.com/jaor/xmobar/blob/master/doc/plugins.org#L1418)?
Thanks again!

guygastineau:
Sure. I will add some docs.  Also, I realized my unicode troubles are not as simple as I thought.  Somehow when I use this unicode that should render in the font I am using doesn't get rendered correctly.  Do you know why?  I would like to fix it before this gets merged.

guygastineau:
This is an example of the unicode breakage in the `QueueReader`.

broken-unicode

In that image I set middot `·` as the `PP` separator from xmonad:

``` haskell
  , ...
  , ppSep = "\x00b7"
  , ...
```

Xmobar prints that character as well as other font icons, as you can see from my weather section to the right, correctly everywhere except withint the `QueueReader` area of the display.

I am writing to it from xmobar like this by overriding my `ppOutput`:

``` haskell

dynamicLogWithPP pp { ppOutput = STM.atomically . STM.writeTQueue q }
```

Unicode works in that section when I use `StdinReader`, and as you can see my ``xpm icons are still working.  I am just stuck figuring this out...

liskin:
If you're using xmonad before https://github.com/xmonad/xmonad-contrib/pull/471 got merged (and there was unfortunately no release after that), then that's a known bug in xmonad.

liskin:
For a quick fix, apply https://hackage.haskell.org/package/utf8-string-1.0.2/docs/Codec-Binary-UTF8-String.html#v:decodeString before writing to the queue. And then don't forget to drop this when you upgrade xmonad.

guygastineau:
@liskin you are so very helpful! <3  I just decided to vendor `xmonad` and `xmonad-contrib` in my project, since I might want to hack on them too anyway (I am really enjoying how flexible cabal the tool feels after so long with stack.  I am building from master right now, so I guess that will fix it for me.

I guess that means I just need to add some documentation to this PR then.  thanks again.

guygastineau:
Okay, I think I made some alright docs for `QueueReader a`.  It needs proof-reading for language and review for typos in the extended example again before merging.

guygastineau:
The example is a little long, but I thought it was better than just regurgitating the contrived example included with the source code.

doc/plugins.org
```diff
@@ -1436,6 +1436,50 @@ choice.
...
+
+When invoking xmobar from other Haskell code it can be easier and
+more performant to use shared memory.  The following plugins leverage
+=Control.Concurrent.STM= to realize these gains for xmobar.
+
+*** =QueueReader (TQueue a) (a -> String) String=
+
+ - Display data from a Haskell =TQueue a=.
+
+ - This plugin is only usefule if you are running xmobar from another
```
jaor, line 1449: typo: useful

jaor (changes requested):
thanks for the docs, they look good (and i like the extended
example) i just pointed out a typo below.

[edit: misunderstanding on my part about docs removed]

jaor:
oh, forget about my rant on the placement of the doc section.  it's already where i was asking!!!

doc/plugins.org
guygastineau: I force pushed an ammendment (I figured I would need to rebase if I pushed a commit just to fix the typo anyway).  This way, probably no rebase or squash is necessary unless you really love terseness.  Do you do fast forwards?  We don't want to get on Linus' bad side lol

So, you are still okay with the STM section?  I know you retracted when you realized it was placed where you asked.  The way I see I am using the plugin to log from XMonad, but this plugin, can be used to do more general things.  I am not sure if anyone would need it, but TQueues with a custom printer could be used to pass practically anything into xmobar.  Anyway, I agree at least that it makes sense having it near the HandleReader as they are very similar.

jaor: yes, i think we're fine right now... and i usually prefer ff, yes, although it's a big deal... i'll let the CI run and merge!

PR #573, opened 2021-09-26
fix: add a missing dependency

kkweon:
without this, `alsa-core` can't be built and `xmobar` build will fail.

I had a build error and found a solution here https://wiki.haskell.org/ALSA


jaor:
these dependencies are optional, and already metioned in https://github.com/jaor/xmobar/blob/master/doc/compiling.org#L96 and https://github.com/jaor/xmobar/blob/master/doc/compiling.org#L101

Issue #571, opened 2021-09-17
QueueReader improvements


jsoo1:
I started using the new QueueReader and it works pretty well save for a few issues:

- The documentation for the plugin could document what the item transformation function is for (the type signature is pretty clear, but still).
- The documentation should document what modules are being imported.
- The documentation should make clear that the alias for use in the template is defined when using Xmobar.QueueReader
- Using xmobar from other haskell code seems to have problems with the dbus module. Perhaps a more direct ipc can be created since using the QueueReader should be done from other Haskell code.
- Actions either don't work or I misunderstand how they should work when rendered directly from other haskell code.
- pp{Hidden,Urgent,Current,Visible} seem to not be updating immediately when I switch workspaces. Perhaps this is a lazy IO bug? Perhaps using [flushTQueue](https://hackage.haskell.org/package/stm-2.5.0.1/docs/Control-Concurrent-STM-TQueue.html#v:flushTQueue) could be used instead of readTQueue

Thanks!  So far it otherwise does exactly what it says on the label :)

guygastineau:
> The documentation for the plugin could document what the item transformation function is for (the type signature is pretty clear, but still).

Yeah, I guess I thought it was obvious, but that is the wrong mindset for the docs to have.  I should have added haddock comments for the fields of QueueReader too.

> Using xmobar from other haskell code seems to have problems with the dbus module. Perhaps a more direct ipc can be created since using the QueueReader should be done from other Haskell code.

This seems like an issue with the dubs plugin.  Nice catch though.

> Actions either don't work or I misunderstand how they should work when rendered directly from other haskell code.

I don't know what that means.

guygastineau:
> pp{Hidden,Urgent,Current,Visible} seem to not be updating immediately when I switch workspaces. Perhaps this is a lazy IO bug? Perhaps using flushTQueue could be used instead of readTQueue

Hmm, I am not experiencing this.if anything is in the queue it should be read by the plugin.  It is really hard to know why this is happening without reproduction.  Maybe you can share the driver code?

I `flushTQueue`  doesn't retry, and it will produce an empty list.  That require state to keep the latest string around.  Reading the internal code for `flushTQueue` I don't think it will force a write.  It just sticks the list together from the FIFO and sets the FIFO's lists to `[]`.

Also, readTQueue should block itself and retry indefinitely as a thread, so it should read new insertions in the queue as soon as the scheduler prioritizes it.  That assumes it is not somehow running behind xmonad.

[...]

PR #574, opened 2021-09-26
Ipc improvements when used from other Haskell code

jsoo1:
I think this needs some documentation and thought but I can confirm that I can send signals to xmobar without dbus by keeping around the TMVar.

I mentioned this in #571

### DIFF
--- a/src/Xmobar/Config/Parse.hs
+++ b/src/Xmobar/Config/Parse.hs
@@ -57,7 +57,7 @@ parseConfig defaultConfig =
         x <- perms
         eof
         s <- getState
-        return (x,s)
+        return (x (SignalChan Nothing),s)
 
       perms = permute $ Config
               <$?> pFont <|?> pFontList <|?> pWmClass <|?> pWmName
--- a/src/Xmobar/Config/Parse.hs
+++ b/src/Xmobar/Config/Parse.hs
@@ -106,7 +106,8 @@ parseConfig defaultConfig =
       pAlpha = readField alpha "alpha"
       pVerbose = readField verbose "verbose"
 
-      pSignal = field signal "signal" $ fail "use default signal"
+      pSignal = field signal "signal" $
+        fail "signal is meant for use with Xmobar as a library.\n It is not meant for use in the configuration file."
 
       pCommands = field commands "commands" readCommands
 
--- a/src/Xmobar/Config/Parse.hs
+++ b/src/Xmobar/Config/Parse.hs
@@ -57,7 +57,8 @@ parseConfig defaultConfig =
         x <- perms
         eof
         s <- getState
-        return (x (SignalChan Nothing),s)
+        let sig = signal defaultConfig
+        return (x sig ,s)
 
       perms = permute $ Config
               <$?> pFont <|?> pFontList <|?> pWmClass <|?> pWmName
--- a/src/Xmobar/Config/Parse.hs
+++ b/src/Xmobar/Config/Parse.hs
@@ -57,8 +57,7 @@ parseConfig defaultConfig =
         x <- perms
         eof
         s <- getState
-        let sig = signal defaultConfig
-        return (x sig ,s)
+        return (x, s)
 
       perms = permute $ Config
               <$?> pFont <|?> pFontList <|?> pWmClass <|?> pWmName
@@ -69,7 +68,7 @@ parseConfig defaultConfig =
               <|?> pAllDesktops <|?> pOverrideRedirect <|?> pPickBroadest
               <|?> pLowerOnStart <|?> pPersistent <|?> pIconRoot
               <|?> pCommands <|?> pSepChar <|?> pAlignSep <|?> pTemplate
-              <|?> pVerbose
+              <|?> pVerbose <|?> pSignal
 
       fields    = [ "font", "additionalFonts","bgColor", "fgColor"
                   , "wmClass", "wmName", "sepChar"
@@ -77,7 +76,7 @@ parseConfig defaultConfig =
                   , "position" , "textOffset", "textOffsets", "iconOffset"
                   , "allDesktops", "overrideRedirect", "pickBroadest"
                   , "hideOnStart", "lowerOnStart", "persistent", "iconRoot"
-                  , "alpha", "commands", "verbose"
+                  , "alpha", "commands", "verbose", "signal"
                   ]
 
       pFont = strField font "font"
@@ -106,6 +105,8 @@ parseConfig defaultConfig =
       pIconRoot = readField iconRoot "iconRoot"
       pAlpha = readField alpha "alpha"
       pVerbose = readField verbose "verbose"
+
+      pSignal = field signal "signal" $ fail "use default signal"
 
       pCommands = field commands "commands" readCommands
 
--- a/src/Xmobar/Config/Types.hs
+++ b/src/Xmobar/Config/Types.hs
@@ -99,7 +99,7 @@ data Border = NoBorder
 newtype SignalChan = SignalChan { unSignalChan :: Maybe (STM.TMVar SignalType) }
 
 instance Read SignalChan where
-  readsPrec _ s = [ (SignalChan Nothing, s) ]
+  readsPrec _ _ = fail "Trying to read a SignalChan"
 
 instance Show SignalChan where
   show (SignalChan (Just _)) = "SignalChan (Just <tmvar>)"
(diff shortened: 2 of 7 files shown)


src/Xmobar/Config/Types.hs
```diff
@@ -91,3 +95,12 @@ data Border = NoBorder
             | BottomBM Int
             | FullBM Int
               deriving ( Read, Show, Eq )
+
+newtype SignalChan = SignalChan { unSignalChan :: Maybe (STM.TMVar SignalType) }
+
+instance Read SignalChan where
+  readsPrec _ s = [ (SignalChan Nothing, s) ]
```
guygastineau, line 102: Why should there be a real read instance?  Maybe I misunderstand, but it seems like using the channel require access to the `TMVar` which should come from the client?  That is, am I correct in assuming that an empty signal channel read from a config is unusable for calling code?

guygastineau:
Is this about Intra-Process Communication as opposed to Inter-Process Communication?

src/Xmobar/Config/Types.hs
jsoo1: I left the empty Read instance so that users who compile xmobar and spawn it as a separate os process or have a 'standard' configuration file don't end up making a TMVar that they never use.  Really this is for people using xmonad or some other haskell wm who want to use xmobar more like a library.

jsoo1:
> Is this about Intra-Process Communication as opposed to Inter-Process Communication?

ha! i suppose so!

src/Xmobar/Config/Types.hs
jsoo1: An alternative I played with was to accept `Maybe TMVar` in `xmobar` as a second parameter and that removes the need to have it in the config but it doesn't feel quite as nice for the calling code

jsoo1:
i think there is also a possibility of providing some signalling mechanism back to xmonad or some other wm that could potentially be a nicer alternative to `action`s, too. i imagine that should probably be done in conjunction with this for a release.

src/Xmobar/Config/Types.hs
jsoo1: in other words, if you are using xmobar from an xmonad.hs then you can use the signal channel, otherwise, xmobar should create one and do the normal execution with os signal handling setup and such.

guygastineau: The convention that I followed when I made the QueueReader plugin was a failing Read instance which I copied from `HandleReader`:

``` haskell
-- | WARNING: This read instance will throw an exception if used! It is
-- only implemented, because it is required by 'Xmobar.Run` in 'Xmobar.commands'.
instance Read (QueueReader a) where
  -- | Throws an 'error'!
  readsPrec = error "QueueReader: instance is a stub"
```

While this still feels hacky, I think it is better to fail loud and hard if someone tries to add these unreadable plugins to a text config.

guygastineau: Using the failing `Read` also makes it clear that the `STM` object doesn't belong in the config datatype.

src/Xmobar/Config/Types.hs
```diff
@@ -91,3 +95,12 @@ data Border = NoBorder
             | BottomBM Int
             | FullBM Int
               deriving ( Read, Show, Eq )
+
+newtype SignalChan = SignalChan { unSignalChan :: Maybe (STM.TMVar SignalType) }
```
guygastineau, line 99: If the `signal` field isn't in the `Config` type, then this doesn't need to be wrapped in Maybe.  When I look at this without `Maybe _`, I realize this is just a specialization of the type constructor `STM.TMVar a`.  Wouldn't it be better to add a `VarReader` (or some more sensible name) plugin to out `STM` plugin collection that can then be specialized by client code?  That way I can make a `VarReader SignalType` from an `STM.TMVar SignalType` that I make in main.  This can then be put in the `commands` field, and the surface of `Config` never has to change.  This is the point of Plugins as I understand them, and I think this kind of generalization is important to maintain flexibility for users.

guygastineau: Oh, I see, this doesn't display anything (duh), so a `VarReader a` wouldn't be very useful.  You can probably just disregard my review then.

guygastineau: Can `SignalType`s already be sent via the dbus plugin?

jaor:
@jsoo1 you pushed a patch changing alsactl, which seems unrelated to this PR: was that an accident?

jsoo1:
Yes. Pardon me

jaor:
@jsoo1 apologies, i lost track of this PR... are we in a good state for reviewing and, eventually, merging? (well, i'm afraid i pushed other patches and you'd need at least a rebase, but hopefully those other changes are not affecting yours directly) @guygastineau any thoughts from your side?  thanks both!

jsoo1:
No problem! I think I still need to address some comments and it could use a bit of design review still.

src/Xmobar/Config/Types.hs
guygastineau: Hmm, okay it does still seem wrong to me to have read instance for the signal channel do anything but fail, since it should never be used as such.

src/Xmobar/Config/Parse.hs
```diff
@@ -57,7 +57,7 @@ parseConfig defaultConfig =
         x <- perms
         eof
         s <- getState
-        return (x,s)
+        return (x (SignalChan Nothing),s)
```
guygastineau, line 60: This feels like an ad hoc solution that clutters the already dauntingly simple magic happening here.  I think the default value for `signal` should be provided by the `defaultConfig` argument.

guygastineau (changes requested):
I still have the same gripe, but I think I managed not be completely wrong about it this time ;)  I think a dummy read instance for a type that should never be read should fail hard and loud.  I know it needs the instance due to the read constraint on `Config` (which maybe we don't even need?), but reading it `signal` for the config shouldn't happen.  This is an odd case where I am advocating runtime errors.

As noted in my other comment, I think the default value for `signal` should be provided to the `parseConfig` function in the `defaultConfig` parameter instead of being hard-coded into the `parseConf` function.

Other than these two things I don't think I have an issue with this PR.  Good work @jsoo1

jsoo1:
@guygastineau @jaor I updated the error message when trying to parse the `signal` field and when `read`ing a `SignalChan`. What do you think?

jaor:
the error message looks good to me, and the code is reasonable.  could
we see an example of how this would be used in a realistic setting? and
also, i'm curious as to why the same thing cannot be done as
conveniently using dbus signals (sorry if you already explained that and
i've forgotten :))... as shown in PR #499 (which i should really
clean-up and merge one of these days), the latter is quite simple and
maybe could be reused somehow to the same effect (but i'm not opposed to
this alternative approach, if they're indeed alternatives!).

jsoo1:
> could we see an example of how this would be used in a realistic setting? and also, i'm curious as to why the same thing cannot be done as conveniently using dbus signals (sorry if you already explained that and i've forgotten :))... as shown in PR #499 (which i should really clean-up and merge one of these days), the latter is quite simple and maybe could be reused somehow to the same effect (but i'm not opposed to this alternative approach, if they're indeed alternatives!).

I was using DBus signals before.  Since moving my xmobar configuration in to xmonad.hs, it seemed stringly typed and more cumbersome to use dbus than just sending messages across the TVar.

Here's what I have right now in xmonad.hs:

```hs
startXmobar :: IO (STM.TQueue String, STM.TMVar Xmobar.SignalType, ThreadId)
startXmobar = do
  xmobarQueue <- STM.newTQueueIO
  xmobarSignal <- STM.newEmptyTMVarIO
  conf <- xmobarConf xmobarSignal xmobarQueue
  xmobarProc <- forkIO (Xmobar.xmobar conf)
  pure (xmobarQueue, xmobarSignal, xmobarProc)


-- Later, using the TVar SignalType bound to a keybinding

toggleBar :: STM.TMVar Xmobar.SignalType -> X ()
toggleBar xmobarSignal = do
  broadcastMessage ToggleStruts
  io (STM.atomically (STM.putTMVar xmobarSignal (Xmobar.Toggle 0)))
  -- ^^^^ This is much nicer than constructing and sending a dbus message
  -- (not to mention, less error prone and faster)
  refresh
```

Here's my xmonad.hs where i've been trying  some new things out.

https://github.com/jsoo1/dotfiles/blob/release/xmonad/xmonad.hs#L71
https://github.com/jsoo1/dotfiles/blob/release/xmonad/xmonad.hs#L106

jaor:
merged.  thanks for your patience!

jsoo1:
No problem! Thank you!

PR #575, opened 2021-10-04
add reading memory specific for freebsd

michal800106:
In freebsd /proc/memoryinfo is absent so we should use sysctl for
obtaining info about stats of memory.


jaor:
looks good, even though i don't have a freebsd system nearby to
properly test it.  merged, thanks!

PR #578, opened 2021-10-11
add top for freebsd proc

michal800106:
In FreeBSD /proc/pid/stat is missing we should use for top procstat
library.


jaor:
i don't understand why the CI jobs aren't working... however, there's definitely compilation errors here on linux:

```
src/Xmobar/Plugins/Monitors/Top/Linux.hs:92:3: error:
    • No instance for (Num (IO Float))
        arising from a use of ‘fromIntegral’
    • In the first argument of ‘(/)’, namely ‘fromIntegral cr’
      In a stmt of a 'do' block: fromIntegral cr / 100
      In the expression:
        do cr <- getSysVar ClockTick
           fromIntegral cr / 100
   |
92 |   fromIntegral cr / 100
   |   ^^^^^^^^^^^^^^^

src/Xmobar/Plugins/Monitors/Top/Linux.hs:92:3: error:
    • No instance for (Fractional (IO Float)) arising from a use of ‘/’
    • In a stmt of a 'do' block: fromIntegral cr / 100
      In the expression:
        do cr <- getSysVar ClockTick
           fromIntegral cr / 100
      In an equation for ‘scale’:
          scale
            = do cr <- getSysVar ClockTick
                 fromIntegral cr / 100
   |
92 |   fromIntegral cr / 100
   |   ^^^^^^^^^^^^^^^^^^^^^
```

michal800106:
I've done both problems. I added apt-get update for CI, and missing return $ for scale.

michal800106:
Thanks for merging. I will check if xmobar is accessible from packages in freebsd because as I know there is a trend for installing haskell stuff from ports or manually by using cabal.

 Additionally I was trying to run tests for freebsd and I found one working way by using self-hosted and github-act-runner. But the issue is that I run it on my computer and it is not always running.
 Github doesn't want to make freebsd image because support for dotnet.
 There is also approach for running this tests by macos image and running nested freebsd in virtualbox. But when I run it I receive timeout after half hour.

 The example of self-hosted run:
https://github.com/michal800106/xmobar/actions/runs/1354197426

Maybe for this moment will be running tests in forks and using locally this self-hosted approach.
But maybe there is better solution for this which I don't know...

michal800106:
I added https://github.com/michal800106/xmobar/blob/test_runner/.cirrus.yml 
And result for first success run (without cache):
https://cirrus-ci.com/task/6248737279311872

For me looks it quite good... :)
Can I make new PR with it? I am worrying only about dual solution: separated for ubuntu and freebsd.
But maybe it will work smoothly.

jaor:
On Mon, Oct 18 2021, Michał Zielonka wrote:

> I added https://github.com/michal800106/xmobar/blob/test_runner/.cirrus.yml 
> And result for first success run (without cache):
> https://cirrus-ci.com/task/6248737279311872
>
> For me looks it quite good... :) Can I make new PR with it?

i don't see why not! :) let's try, please.

> I am worrying only about dual solution: separated for ubuntu and
> freebsd.  But maybe it will work smoothly.

in what sense? i think i can configure the main repo to wait on reports
from external CIs like cirrus... or are you worried about something
else?

PR #580, opened 2021-10-19
add cirrus freebsd test

michal800106:
For start cirrus please use:
https://cirrus-ci.org/guide/quick-start/
choose public repositories plan and add only xmobar as observed by cirrus.

Also here is addes small fix for dividing by zero when cpu usage is calculated


jaor:
oh, perfect! thanks a lot for your great work!

PR #586, opened 2021-10-30
dynamic pads <lpad=X> and <rpad=X> support

tulth:
Relating to issue https://github.com/jaor/xmobar/issues/583

Looking for feedback.  I am not an expert on the parsing side of Haskell so that should get extra review.

This PR adds a capability to add dynamic padding to either the right or left side of xmobar.  Example use: making room for system trays.

(unsafe)Stdin can include, for example, <rpad=123> to reserve 123 pixels on the right hand of xmobar.  The xmobar window is shrunk on the right by this amount. 

Note that dynamic padding by <pad/lpad=X> is on top of any padding requested by position TopP BottomP.  I believe this is the right strategy because if someone is using TopP to get a bar sized just so, they still may want dynamic padding in addition to the TopP/BottomP requested amount.

The rpad/lpad are stored as new keys in the Config record.  I welcome suggestions on whether it should be structured differently.

I also added some additional Show derived capabilities to Widgets/etc to aid in debug.

### DIFF
--- a/src/Xmobar/App/Config.hs
+++ b/src/Xmobar/App/Config.hs
@@ -60,6 +60,8 @@ defaultConfig =
            , commands = [ Run $ Date "%a %b %_d %Y * %H:%M:%S" "theDate" 10
                         , Run StdinReader]
            , sepChar = "%"
+           , dynLPad = 0
+           , dynRPad = 0
            , alignSep = "}{"
            , template = "%StdinReader% }{ " ++
                         "<fc=#00FF00>%uname%</fc> * <fc=#FF0000>%theDate%</fc>"
--- a/src/Xmobar/App/EventLoop.hs
+++ b/src/Xmobar/App/EventLoop.hs
@@ -38,8 +38,9 @@ import Control.Concurrent.STM
 import Control.Exception (bracket_, handle, SomeException(..))
 import Data.Bits
 import Data.Map hiding (foldr, map, filter)
-import Data.Maybe (fromJust, isJust)
+import Data.Maybe (fromJust, isJust, mapMaybe)
 import qualified Data.List.NonEmpty as NE
+import qualified Data.Bifunctor
 
 import Xmobar.System.Signal
 import Xmobar.Config.Types
@@ -164,8 +165,13 @@ eventLoop tv xc@(XConf d r w fs vos is cfg) as signal = do
             xc' <- updateCache d w is (iconRoot cfg) str >>=
                      \c -> return xc { iconS = c }
             as' <- updateActions xc r str
-            runX xc' $ drawInWin r str
-            eventLoop tv xc' as' signal
+            let possiblePads = extractDynPad str
+                cfg' = maybe cfg (\pads -> cfg { dynLPad=fst pads
+                                               , dynRPad=snd pads}) possiblePads
+                xc'' = xc' {config=cfg'}
+            r' <- repositionWin d w (NE.head fs) cfg'
+            runX xc'' $ drawInWin r' str
+            eventLoop tv xc'' as' signal
 
          Reposition ->
             reposWindow cfg
@@ -265,6 +271,7 @@ updateActions conf (Rectangle _ _ wid _) ~[left,center,right] = do
       iconW i = maybe 0 Bitmap.width (lookup i $ iconS conf)
       getCoords (Text s,_,i,a) = textWidth d (safeIndex fs i) s >>= \tw -> return (a, 0, fi tw)
       getCoords (Icon s,_,_,a) = return (a, 0, fi $ iconW s)
+      getCoords (Pad _,_,_,a) = return (a, 0, 0)
       partCoord off xs = map (\(a, x, x') -> (fromJust a, x, x')) $
                          filter (\(a, _,_) -> isJust a) $
                          scanl (\(_,_,x') (a,_,w') -> (a, x', x' + w'))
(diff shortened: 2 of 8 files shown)


jaor:
this looks interesting. but a thing that i'm not sure about is that, as
far as i can tell, we're letting people writing a pad tag anywhere in
their templates, aren't we?  but this only makes sense as the output fo
Stdin: what happens if i put one of these paddings in the middle of my
template, or two? are we signalling an error? if we cannot find a
reasonable answer to those questions (and i don't see any natural one),
my instinct is that this functionality does not belong to the template
markup, but, rather, to some kind of signalling...

tulth:
> this looks interesting. but a thing that i'm not sure about is that, as far as i can tell, we're letting people writing a pad tag anywhere in their templates, aren't we? but this only makes sense as the output fo Stdin: what happens if i put one of these paddings in the middle of my template, or two? are we signalling an error? if we cannot find a reasonable answer to those questions (and i don't see any natural one), my instinct is that this functionality does not belong to the template markup, but, rather, to some kind of signalling...

Just to make sure I understand, are you talking about the changes to Parse.hs?  If so think I can work on removing those.  I agree it should not be in the config file, but only on stdin

liskin:
Hm, this seems like a weird mix of my `<hspace=123/>` suggestion and the original idea of dynamically resizing xmobar. Unfortunately there is no explanation why. :-/

So I wonder: what was wrong with the simple `<hspace=123/>` idea?

If there's something wrong with that, then I agree with @jaor that this might better be signalled outside of the template/markup language. Perhaps using the dbus interface? But please do help us understand what is wrong with just having a markup for horizontal spacing.

jaor:
On Fri, Oct 29 2021, tulth wrote:

> Just to make sure I understand, are you talking about the changes to
> Parse.hs?  If so think I can work on removing those.  I agree it
> should not be in the config file, but only on stdin

yes, that'd be better.  but see the other comments about hspace.

jaor:
On Sat, Oct 30 2021, Tomáš Janoušek wrote:

> Hm, this seems like a weird mix of my `<hspace=123/>` suggestion and
> the original idea of dynamically resizing xmobar. Unfortunately there
> is no explanation why. :-/
>
> So I wonder: what was wrong with the simple `<hspace=123/>` idea?

i suppose that if we had `<hspace=...>`, for the trayer case the default
template wouldn't have any hspace and it'd be produced on demand by
Stdin or the XProp monitors.  but for that to work as a gap, Stdin would
need to appear in the template either first or last (for left and right
padding of the whole bar).

but i think that problem can be solved by instances of NamedPropertyLog
which would be set to the desired `<hspace>` on demand from xmonad,
leaving Stdin alone (which shouldn't be used anyway, since we have
XPropertyLog and friends, but that's a tangent).

that should work and be cleaner and more reusable than a adhoc padding
limited to Stdin, right?

liskin:
> but i think that problem can be solved by instances of NamedPropertyLog which would be set to the desired `<hspace>` on demand from xmonad, leaving Stdin alone (which shouldn't be used anyway, since we have XPropertyLog and friends, but that's a tangent).
> that should work and be cleaner and more reusable than a adhoc padding limited to Stdin, right?

Yes, that's how I envisioned this.

On the xmonad side, having an [event hook](https://github.com/jaor/xmobar/issues/583#issuecomment-951317168) put additional content into the main XMonadLog property (or stdin) isn't entirely straightforward and requires an extra IORef or ExtensibleState and explicit invocation of the logHook, whereas just writing into an extra property is much easier.

tulth:
Ok I have a 2 questions and a use case I would like to get a reaction to.

### Q1
Should this support
- padding with blank areas or
- actual xmobar window resizing
- or both?

I like resizing because it is robust against mistakes in the window layering.  In other words, if the xmobar is ever considered to be on top of the tray, xmobar could draw on top of the tray with spacing.

However, I can see the argument for ```<hspace=X>``` in general for this use and other uses.

### Q2
You requested I change the padding to instead use the XMonadLog method of input instead of stdin. Question: I think because XMonadLog messages still have to be parsed using Parsers.hs so unless there are different parsers, the current setup will not prevent the user from sending the same commands to stdin as XMonadLog, correct?

But I think it does remove the need to say which side its on, as long as the user is willing to use two different XPropertyLogs.

### Use case
So putting it in context for a use cases,  just using a single xmobar to keep it simple, and in this case the xmonad prop log goes on the left and the padding for the tray is on the right:

In xmonad.hs
```haskell
main = do
  mySB <- statusBarPipe "xmobar" (pure myPP)
  xmonad $ withSB mySB myConf
-- where myConf includes myTrayEventHook

-- an event hook to report tray size areas, something like this:
myTrayEventHook :: Event -> X All
myTrayEventHook (PropertyEvent { ev_window = w }) = do
    whenX (runQuery checkDock w <&&> (runQuery panelQuery w)) (
      withDisplay $ \d -> withWindowAttributes d w $ \wa -> do
        let pixelPad = (fromIntegral $ wa_width wa)
        xmonadPropLog' "_XMONAD_PAD" ("<rpad=" ++ show pixelPad ++ ">") -- or hspace=?
      )
    return (All True)
myTrayEventHook _ _ = return (All True)

panelQuery :: Query Bool
panelQuery = appName =? "panel"
```

in xmobarrc:
```
Config { ...
       , commands = [ Run XPropertyLog "_XMONAD_LOG", ... ]
       , commands = [ Run XPropertyLog "_XMONAD_PAD", ... ]
       , template = "%_XMONAD_LOG% }{ ... %_XMONAD_PAD%"
       }
```

liskin:
You're right that with my `<hspace=…>` suggestion, layering will be an issue. It's more or less a solved problem in xmonad-contrib, but if this should be useful in general, then perhaps dynamic margins (= resizing) are indeed useful.

I'll leave it up to @jaor to answer the other questions. But if this was up to me, I'd consider implementing both: a simple `<hspace=…>` markup and a dbus API for dynamic left/right margins. Simply because the former is so simple and can be useful elsewhere.

jaor:
On Sun, Oct 31 2021, Tomáš Janoušek wrote:

> I'll leave it up to @jaor to answer the other questions. But if this
> was up to me, I'd consider implementing both: a simple `<hspace=…>`
> markup and a dbus API for dynamic left/right margins. Simply because
> the former is so simple and can be useful elsewhere.

i tend to agree. resizing sounds like a departure from the usual markup
semantics, which never affect the global geometry of the bar (monitors
can print strings of different lenghts, but never affect the bar
itself), and sounds a bit harder to implement at that level, while
adding dbus signals is pretty easy (see PR #499 for an example).

so, in an ideal world, i'd like to have a static hspace and a resizing
dbus signal :) but i can settle for other combinations if their
implementation keeps the code simple/orthogonal.

tulth:
closing in favor of the hspace approach

PR #593, opened 2021-12-22
Fix BatteryP example configuration

andreasdoll:
Add a forgotten comma in example configuration to make it syntactically
correct.


jaor:
merged. many thanks for taking the time!

PR #603, opened 2022-01-31
Text output

pavel-the-best:


examples/xmobar.config
```diff
@@ -15,6 +15,7 @@ Config { font = "-misc-fixed-*-*-*-*-10-*-*-*-*-*-*-*"
        , iconRoot = "."
        , allDesktops = True
        , overrideRedirect = True
+       , textOutputColors = Just Ansi
```
jaor, line 18: the code looks good, and i'm merging it later... but here in the configuration file i'd rather have Ansi (or Pango), without the `Just`, which is a bit of an abstraction leak, and to make things easier for users not knowing haskell.

PR #605, opened 2022-02-05
wip: full implementation of swaybar-protocol

jaor:
New text output format, `Swaybar`, that emits and reads JSON according
to swaybar-protocol.  We just translate our template markup to swaybar
specs, when available.

  - [x] Colours
  - [x] Clicks via `<action/>`
  - [x] Colour transparency
  - [x] Boxes around text
  - [x] Multiple fonts


src/Xmobar/Config/Types.hs
```diff
@@ -98,7 +98,7 @@ data Border = NoBorder
             | FullBM Int
               deriving ( Read, Show, Eq )
 
-data TextOutputFormat = Plain | Ansi | Pango deriving ( Read, Show, Eq )
+data TextOutputFormat = Plain | Ansi | Pango | Swaybar deriving (Read, Show, Eq)
```
liskin, line 101: This should be added to `--help` (`src/Xmobar/App/Opts.hs`) as well.

xmobar.cabal
```diff
@@ -184,24 +186,25 @@ library
...
-      transformers,
-      X11 >= 1.6.1,
-      mtl >= 2.1 && < 2.3,
-      parsec == 3.1.*,
-      parsec-numbers >= 0.1.0,
-      stm >= 2.3 && < 2.6,
-      extensible-exceptions == 0.1.*,
-      async
+                  X11 >= 1.6.1,
+                  aeson >= 2.0.3.0,
```
liskin, line 190: It builds with `aeson-1.4.7.1` just fine, although I have no way to test whether it works 100%. Tried `-TSwaybar` and it doesn't crash and the output looks reasonable to the naked eye. :-)

PR #608, opened 2022-02-13
Use nh' for displaying the bar on the bottom of the screen. Fixes #555

### LINKED ISSUE #555: Bar on bottom is broken.

exorcist365:
The title says it all.


jaor:
merged, thanks!

PR #609, opened 2022-02-17
Fix memory leak in drawInWin

liskin:
In f8c835a33a7a, I flipped the discard flag to XSync to False on a false assumption that it may discard events from under the eventer thread (since renamed to handleXEvent). This can't happen—the eventer thread and the main loop do not share a Display connection, they have two separate ones. Turns out, the main loop doesn't read/process any events from its Display connection, which is why it was necessary to discard them once in a while.

The fix restores that discarding, adds a comment to explain why that discarding should stay, and just to make things a bit cleaner, also prevents some of those events from being emitted in the first place: by configuring the graphics context that we don't want any exposure events (<https://tronche.com/gui/x/xlib/events/exposure/graphics-expose-and-no-expose.html>).

Fixes: f8c835a33a7a ("Fix delayed reaction to USR1/2 signals")

---

plus a refactor to improve naming of event/signal loop functions (f50453a)

### DIFF
--- a/src/Xmobar/X11/Draw.hs
+++ b/src/Xmobar/X11/Draw.hs
@@ -70,6 +70,7 @@ drawInWin wr@(Rectangle _ _ wid ht) ~[left,center,right] = do
 #endif
   withColors d [bgColor c, borderColor c] $ \[bgcolor, bdcolor] -> do
     gc <- liftIO $ createGC  d w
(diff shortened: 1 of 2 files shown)


jaor:
merged! special thanks for taking the time to thoroughly document the
rationale behind the fixes.

PR #611, opened 2022-02-20
Documentation: QueueReader docs improvements

liskin:
#### [Documentation: Drop unused fragment in QueueReader example](../commit/80b8030505699e9a734ea83423d526d97e0684ec)

Related: https://github.com/jaor/xmobar/issues/571

#### [Documentation: More robust QueueReader example](../commit/7cca823651a4a33e0c99d4062a382704f7904572)

1. Call `initThreads` early because there's no guarantee the xmobar
   thread will do this before xmonad makes any Xlib calls (although it's
   likely as xmonad startup is probably more expensive).

2. Use `forkOS` to fail early if running in a non-threaded RTS. Also,
   xmobar's own main thread would normally be a bound thread, so this
   may (likely non-observably) result in more consistent behaviour.

Related: https://github.com/jaor/xmobar/issues/571

#### [Documentation: Note the need for threaded RTS with QueueReader & xmonad](../commit/efa9e4c541695e725d39da001592c9613acc4cae)

Related: https://github.com/jaor/xmobar/issues/571

### DIFF
--- a/doc/window-managers.org
+++ b/doc/window-managers.org
@@ -208,19 +208,6 @@ choice.
           logWorkspacesToQueue :: STM.TQueue String -> X ()
           logWorkspacesToQueue q =
             dynamicLogWithPP def { ppOutput = STM.atomically . STM.writeTQueue q }
-            where
-              -- Manage the PrettyPrinting configuration here.
-              ppLayout' :: String -> String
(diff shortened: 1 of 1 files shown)


jaor:
thanks!

PR #618, opened 2022-03-29
Fix MultiCoreTemp's temperature file finding logic

### LINKED ISSUE #616: MultiCoreTemp plugin fails when hwmon directory is numbered higher than 9
If the CPU temperatures are located in a hwmon folder numbered higher than 9 (e.g. `/sys/bus/platform/devices/coretemp.0/hwmon/hwmon10`), xmobar will fail to find any CPU temperature files and return an empty list from the `hwmonPaths` function. This propagates through until it throws `Prelude.maximum: empty list`, which is displayed on the bar.

From a bit of digging, this happens because the list `[0..9]` is hardcoded [here](https://github.com/jaor/xmobar/blob/master/src/Xmobar/Plugins/Monitors/MultiCoreTemp.hs#L92-L93=).

This could probably be solved by just listing the directory and taking the matching directories within instead of iterating over a hard-coded list, e.g. 

```haskell
filter ("hwmon" `isPrefixOf`) <$> listDirectory "/sys/bus/platform/devices/coretemp.0/hwmon"
```

However, perhaps there was some sort of performance reason for the hardcoded list I'm unaware of. If not and the above solution seems satisfactory, I'll happily open a pull request.

This hardcoded `[0..9]` list is also present in the `coretempPath` and `corePaths` functions as well, which might also be an issue.

robwhitaker:
Instead of searching for a fixed set of files and directories (numbered
0-9), which would miss anything above 9, it now searches the relevant
directories for files matching the right pattern, regardless of number.

Fixes #616.

### DIFF
--- a/src/Xmobar/Plugins/Monitors/MultiCoreTemp.hs
+++ b/src/Xmobar/Plugins/Monitors/MultiCoreTemp.hs
@@ -122,7 +122,7 @@ corePaths :: Maybe String -> IO [String]
 corePaths s = do ps <- case s of
                         Just pth -> return [pth]
                         _ -> hwmonPaths
-                 cps <- concat <$> traverse (flip getMatchingPathsInDir corePathMatcher) ps
+                 cps <- concat <$> traverse (`getMatchingPathsInDir` corePathMatcher) ps
                  ls <- filterM doesFileExist cps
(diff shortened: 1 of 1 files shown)


jaor (approved):
LGTM! :thumbsup:

PR #620, opened 2022-04-03
Template datatype

### LINKED ISSUE #588: Formalize template grammar?
I did something rash on my quest to use xmobar more as a library than a binary.  My goal after #574 (ipc signals with STM) was to accomplish `action`s with haskell IO actions as well as straight up spawning shells.  I started working on it and realized I would want to have a datatype to parse the template into instead of `String`. This turned out to be a lot of work (though I am quite close to a working version). [contributing.org](https://github.com/jaor/xmobar/blob/master/contributing.org#contributing-changespatches-to-xmobar) says to discuss non-trivial changes with the community before doing anything else. I am sorry I did not raise this earlier.

My question is: Is the community interested in a more formal grammar and datatype for the template to replace the big `String` we have now? I made a `Bar` type and adjusted the template parser. Along with the aforementioned possibility of `IO` `action` segments, other improvements include a test suite for the template parser, better error messages for debugging, a cleaner template parser, and simplified logic in the event loop.  At some point, the new datatype can also be documented as a grammar for the template.

jsoo1:
Closes #588 

Pardon the commit history, it's a bit of a mess. I went through quite a few revisions of this.  The goal here is to support more xmobar configuration as a library.  In this case, you get first-class datatypes to define the bar. This is a nice improvement over writing the pseudo-xml configuration we have now.

It also formalizes the template grammar and improves the parser.

The primary change here is replacing the old `String` that held the bar with a datatype called a `Bar a`. The `Bar` is parameterized by a type so that you can `fmap`, `traverse`, etc over it. The nice thing here is that when `xmobar` starts, it parses the `Bar` configuration from the template string into `Bar RunnableWidget`. Then when the main loop starts and the various widgets have been spawned, it becomes `Bar Running`.  

Not only does this make clear the states that `xmobar` can be in, it also communicates when and where parsing happens.  Because of that, it is clear when and where parse errors should be reported to the user. I believe this should help with some of the difficulties debugging template configurations.

I should also say that I hope not to have introduced any parsing problems, so any user of the string xml templating style should not be effected.

I have not formally benchmarked this change, but on average it appears to use ~100M of resident space which is on-par with the previous revision. There is one slow space-leak I have not yet tracked down which leaks a few mb every day or so.


jsoo1:
For an example of usage as a library, here are some bar segments in my `xmonad` configuration: https://git.sr.ht/~jsoo/dotfiles/tree/release/item/xmonad/xmonad.hs#L284

jaor:
On Sun, Apr 03 2022, John Soo wrote:

> Closes #588
>
> Pardon the commit history, it's a bit of a mess.

Would it be okay to rebase and squash them to a single commit? I'd
prefer it unless you feel strongly against for some reason.

> I went through quite a few revisions of this. The goal here is to
> support more xmobar configuration as a library. In this case, you get
> first-class datatypes to define the bar. This is a nice improvement
> over writing the pseudo-xml configuration we have now.

It'd be nice, if possible, to have some documentation or example
included in the PR, perhaps in the "writing your own pluging" document.
Do you think that'd be possible?

> It also formalizes the template grammar and improves the parser.

Could you at least comment on that formalization in the PR or, even
better, in the squashed commit description?

Thanks a lot!
Cheers,
jao
-- 
I always pass on good advice. It's the only thing to do with it. It is
never any use to oneself.
 -Oscar Wilde

jaor:
moved to https://codeberg.org/jao/xmobar/pulls/620

PR #623, opened 2022-04-12
make small refactoring of load linux code

michal800106:
Add Data.Factor.(<&>) and use it for parsing content of file.


jaor:
thanks!

to be honest, i'm not a big fan of this refactoring.  it transforms
code that i can read and understand immediately, to one that i can
only guess at without looking at the documentation of `<&>`. so, in my
humble opinion, it's just obfuscation for the sake of saving a line of
code.

`<rant>` i must admit that i profoundly dislike haskell code full of
operators without obvious meaning.  makes me think of perl. `</rant>`

all that said, if other contributors would chime in and share their
opinion, i don't discard at all that i'm just being ignorant and that
most haskellers know about `<&>` and find it as natural as `>>=`.  in
which case i'll merge without troubles :)

liskin:
> all that said, if other contributors would chime in and share their
> opinion, i don't discard at all that i'm just being ignorant and that
> most haskellers know about `<&>` and find it as natural as `>>=`. in
> which case i'll merge without troubles :)

Personally I do find `<&>` as natural as `<$>` and `<*>`, which is probably about as natural as `>>=`, but not quite. I do, however, tend to only use it when followed by a lambda (`f <&> \x -> x + 1`) as it lets me avoid parentheses (especially helpful when the lambda needs a line break). Without the lambda, `<$>` is often more readable, as it follows the natural function application order from right to left.

So this entire patch can be replaced by a single-line change to `Result . parseLoadAvgs <$> B.readFile file` … :-)

michal800106:
:) honestly I used ```(<&>)``` because hlint makes this suggestion and complains about using ` >>= return .`  Personally for me the best solution will be proposed by @liskin and I've pushed version with ```(<$>)```.

Issue #625, opened 2022-04-15
Battery power calculation gives wrong unit


pguenth:
On the laptop I use the power_supply driver only reports `current_now` and not `power_now`. I can't find a definite source for the unit of `power_now`, but `current_now` seems to be given in uA as it is stated in the kernel source code:

Line 22 in [this](https://github.com/torvalds/linux/blob/master/include/linux/power_supply.h) file: 
> All voltages, currents, charges, energies, time and temperatures in uV, µA, µAh, µWh, seconds and tenths of degree Celsius unless otherwise stated.

Looking at the [xmobar source code](https://github.com/jaor/xmobar/blob/master/src/Xmobar/Plugins/Monitors/Batt/Linux.hs) it seems that the only difference in calculating the power shown between systems providing `power_now` and `current_now` is a factor of 10 for some reason that is not clear to me. I try to outline my thought process for this, since I don't have much experience in Haskell I might miss something, so correct me if I'm wrong. In line 89 in the linked file a scaling factor is calculated which is the default (10^6, accounting for the micro-prefix I think) but divided by 10 if a `current_now` file has been found in the function `batteryFiles`. As far as I can see this is the only difference between the two options the kernel may provide and further down the line the value from the file chosen multiplied with this scaling factor is interpreted as watts.

This may be a pretty wild guess but maybe the reason for the different scaling factors is that 

    P[W] = I[A] * U[V] = I[uA]/10^6 * U[V] = I[uA]/10^6 * 10V = (I[uA]/10^5) in Watts

assuming that most laptop batteries have a voltage which is in the order of 10V. Then this calculation gives reasonable values for the power consumption, but obviously not very precise ones. At least on my system the kernel also provides the property `voltage_now` which could simply be used to calculate the correct value (if it exist if this is not the case on all systems, I can't find any information on that topic at the moment).

It is relatively difficult to get a decent comparison with the actual values, but the values I calculate by multiplying `voltage_now` and `current_now` don't seem to add up with those shown in my xmobar although they should be right.

I am happy to provide a pull request to fix this issue if neccessary.

jaor:
Hi Patrick,

I cannot remember where that factor of 10 came from, but your rationale
makes sense.  My kernel does provide a voltage_now file (and it'd be
giving a 8.5 factor rather than 10). If, as in your case and mine, that
file is present, i don't see any reason not to use it and only default
to the 10 guessing otherwise.  So yes, a PR would be most welcome,
many thanks in advance!

PR #626, opened 2022-04-18
Restructuring of battery monitoring on Linux while fixing units

pguenth:
While working on fixing #625 I realized that apart from the wrong calculation of the power value mentioned in the issue that the calculation of the capacity was flawed in the same way. Things seemed to get quite messy so I ended up doing some restructuring on the battery attribute retrieval code. `Maybe` is now used for handling non-existing files, different sets of attributes exposed by the kernel are differentiated using pattern matching and the code for retrieving each quantity is seperate function now.

I hope this is not to considered too much of a change, but if so I will be happy to try and fix the units without the more fundamental changes.

### DIFF
--- a/src/Xmobar/Plugins/Monitors/Batt/Linux.hs
+++ b/src/Xmobar/Plugins/Monitors/Batt/Linux.hs
@@ -23,7 +23,7 @@ import Xmobar.Plugins.Monitors.Batt.Common (BattOpts(..)
 import Control.Monad (unless)
 import Control.Exception (SomeException, handle)
 import System.FilePath ((</>))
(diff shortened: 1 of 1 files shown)


jaor:
this looks essentially good to me, thanks! i've merged it and pushed a
couple of stylistic refactorings on your code in 676c905 (basically,
to get rid of that ugly prefixFiles function, and remove a couple of
unneeded helper functions): please, feel free to complain if you
dislike something in there!

pguenth:
Looks fine to me, you essentially fixed the rough parts where I haven't been able to come up with a better solution due to my limited experience with Haskell. Thank you for reviewing!

PR #628, opened 2022-05-12
Remove with_utf8 option and enable its features by default.

arrowd:
It is not possible to build xmobar without this option enabled:

```
src/Xmobar/Text/SwaybarClicks.hs:34:1: error:
    Could not find module `Data.ByteString.Lazy.UTF8'
    Perhaps you meant
      Data.ByteString.Lazy.Char8 (from bytestring-0.11.3.0)
      Data.ByteString.Lazy (from bytestring-0.11.3.0)
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
34 | import Data.ByteString.Lazy.UTF8 (fromString)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/xmobar.cabal
+++ b/xmobar.cabal
@@ -32,10 +32,6 @@ flag with_xft
   description: Use Xft to render text. UTF-8 support included.
   default: False
 
-flag with_utf8
-  description: With UTF-8 support.
-  default: True
-
 flag with_inotify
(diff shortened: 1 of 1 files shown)


jaor:
merged, thanks.  i've also removed conditional uses of the UTF8 macro.

PR #635, opened 2022-06-05
Add flake.nix

alternateved:
Hello, 

I would like to propose modification to current way of supporting Nix ecosystem. 

The `shell.nix` file that is available in repository seems outdated and does not provide a comfortable way to install xmobar. Also, it explicitly states dependencies, where those could be fetched from cabal file and translated with [cabal2nix](https://github.com/NixOS/cabal2nix).

Currently XMonad is providing [flakes](https://github.com/xmonad/xmonad/blob/master/flake.nix) [files](https://github.com/xmonad/xmonad-contrib/blob/master/flake.nix) which enables easy installation, caching and isolated environment (users do not have to install Haskell dependencies with their system's package manager and mutate their current environment, since everything is contained within flake). I thought it might be helpful to do the same for xmobar.

The example of usage could be seen in my [configuration](https://gitlab.com/alternateved/dotfiles/-/tree/85bdf837c48a0c12c105282290f77db71bbab81f/xmobar/.config/xmobar) where I use flakified xmobar as an input and create my own binary (with `nix build .` command) which is then used from XMonad. This solution also provides easy access to tools like haskell-language-server.

More on practical nix flakes usage could be found [here](https://www.tweag.io/blog/2020-05-25-flakes/) and [here](https://serokell.io/blog/practical-nix-flakes).

Kind wishes
alternateved


jaor:
Hi @alteranteved.  Thanks for this contribution.  I am not a Nix user,
and know very little about it, but what i've seen makes sense and,
honestly, if the xmonad maintainers are doing it too, it cannot be bad
:)

Merging!

